Issue #2, opened 2016-11-08
Switch to new tile provider


jancborchardt:
Following the MapQuest open API shutdown we need to check for a new tile provider.

GNOME Maps uses Mapbox with a community API key over a GNOME proxy: https://mail.gnome.org/archives/commits-list/2016-July/msg07462.html
I contacted them to find out how we can do it: https://twitter.com/jancborchardt/status/796054542144176131

A possible alternative would be the Wikimedia Cartographical project: https://www.mediawiki.org/wiki/Maps#Production_maps_cluster

cc @v1r0x @nextcloud/maps @schiessle

v1r0x:
Definitely a problem. I remember that we switched from vanilla osm aka nominatim to something else for the routing/geocoder stuff, but is there a reason we don't use nominatim as tile server?
Gnome's solution would be great too!
Beside that, I'll look for another solution.

PS: I ordered a new laptop, thus I hope I'll have some more time for maps 😃

jancborchardt:
@v1r0x yeah, sounds great! :) Also, @mattiasb of GNOME Maps replied to the tweets with further info, so we should contact Mapbox: https://twitter.com/jancborchardt/status/796054542144176131 (also cc @freenerd whom I know working there :)

(And cc @Dokhra @e-alfred @thomas-mc-work from the original issue cause we moved the repository here and will continue work here. :)

Henni:
Mapbox would be great, but I haven't found any information on this "community API key" you mentioned @jancborchardt

jancborchardt:
@Henni I’m pretty sure it’s something we need to mail them about. It sounds like mainly a courtesy & community support thing, not an official API key they offer. :)

e-alfred:
Good to see the idea of a maps app isn't dead. Is Openstreetmap also of interest? Offline maps would be quite interesting as well to avoid the problem with an external provider cutting access suddenly.

Maybe this link could be of help:

https://wiki.openstreetmap.org/wiki/Offline_Openstreetmap

jancborchardt:
@e-alfred while having a tile server or offline maps sounds good, it’s hardly doable:

> Planet.osm contains the entire planet. This is a snapshot of the current data, usually from last Wednesday. This is **almost 40 GB compressed**.

Just Germany is 2.5 GB. ;)

Henni:
@e-alfred still... Feel free to open an issue. It might be possible to implement something like offline *area* support, similar to others like Google Maps do it.

[...]

PR #15, opened 2016-12-27
[WIP] Rework maps app

v1r0x:
This is a work in progress PR.

Since the current version relies on an old version of leaflet and a no longer supported tilelayer,  @jancborchardt and I decided to rework the whole app and update it to Leaflet 1.x.

This PR should cover the basic features of a maps app.

See https://github.com/nextcloud/maps/pull/15#issuecomment-324350873 for todo list

cc @nextcloud/maps

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -41,6 +41,7 @@
     var mapController = {
         searchMarker: {},
         map: {},
+        locControl: undefined,
         displaySearchResult: function(result) {
             if(this.searchMarker) this.map.removeLayer(this.searchMarker);
             this.searchMarker = L.marker([result.lat, result.lon]);
@@ -64,6 +65,18 @@
                 maxBounds: new L.LatLngBounds(new L.LatLng(-90, 180), new L.LatLng(90, -180)),
                 layers: [mapQuest]
             });
+            this.locControl = L.control.locate({
+                position: 'topright', // default = topleft
+                drawCircle: false,
+                drawMarker: false,
+                showPopup: false,
+                icon: 'fa fa-map-marker',
+                iconLoading: 'fa fa-spinner fa-spin',
+                strings: {
+                    title: "Get current location"
+                }
+            }).addTo(this.map);
+            this.locControl.start(); // try to get the user's location
         }
     };
 
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "gulp": "^3.9.1",
     "gulp-cli": "^1.4.0",
     "leaflet": "^1.2.0",
+    "leaflet.locatecontrol": "^0.62.0",
     "leaflet.markercluster": "^1.1.0",
     "opening_hours": "^3.5.0"
   }
--- a/templates/content/index.php
+++ b/templates/content/index.php
@@ -19,9 +19,11 @@
  */
 
 style('maps', '../node_modules/leaflet/dist/leaflet');
+style('maps', '../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min');
 script('maps', '../node_modules/leaflet/dist/leaflet');
 script('maps', '../node_modules/leaflet.markercluster/dist/leaflet.markercluster');
 script('maps', '../node_modules/opening_hours/opening_hours');
+script('maps', '../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min');
 ?>
 <div id="search">
     <form id="search-form">


jancborchardt:
Also cc @eneiluj of GpxPod and GpxEdit because we should work together on making this work directly in the one Maps app. :)

jancborchardt:
For the record and to not have it only as one comment in a big issue, here’s the todo list taken from https://github.com/nextcloud/maps/issues/2#issuecomment-313360509 :)

the core scope is to provide an alternative to things like Google Maps, Apple Maps etc. This includes:

Basic core features:
- [ ] Map display
- [ ] Search
- [ ] Elements/POIs being clickable with further information shown
- [ ] Favorites (which can ideally be synced to mobile with Maps.me or OSMand)
- [ ] Routing (for walking, cycling, driving)

Advanced stuff:
- [ ] Ideally even public transport routing, but I know this is hard
- [ ] Integration with other apps like:
  - [ ] Contacts: showing them on the map, and also suggesting them in the search
  - [ ] Contacts: inside contacts with an address, show a little map widget for their location
  - [ ] Files/Gallery: Show photos in their location based on EXIF location data
  - [ ] Calendar: providing the Calendar app with a way to show a little map widget in an event when there is a location set. The other way around, show events which take place in the future and have a location set on that place in the map.
  - [ ] Mail: make location descriptions from mails clickable, or also show a small map widget like that
- [ ] In the future GPX track overlay, possibility to edit

e-alfred:
Postgresql compatibility should be on the todo list too as mentioned here, if changes are necessary: https://github.com/nextcloud/maps/issues/19

v1r0x:
> Routing (for walking, cycling, driving)

@jancborchardt I tried to add the [leaflet-routing-machine](https://github.com/perliedman/leaflet-routing-machine) to the app, but the builtin routing service (osrm) is "[down](https://github.com/Project-OSRM/osrm-backend/wiki/Demo-server)". The mention Mapbox as alternative, but it requires an API key. Since there is no completely free routing service we should decide which one we want to use. IMO there are three options:
1. Mapbox (Supports at least car, bike and walking)
2. Mapzen (Supports at least car, bike, walking and transit; Geocoding, Autocomplete Search)
3. GraphHopper (Supports at least car, bike and walking)

I think mapzen has the most features and a good free plan. GraphHopper looks good as well. Not sure about the pricing plan of mapbox ("credit" based). Any thoughts?

e-alfred:
@v1r0x Mapbox is also used by GNOME Maps since Mapquest shut down their free access. An API key shouldn't be a huge deal, the weather app also requires one to work (pulling data from Openweathermap).

Mapzen sounds surely better because it has more features that are useful for a map app after all. Transit data would be a good addition too.

One important issue I can see would be a good way to make transition to a new backend easier so the Mapquest debacle can be avoided if a provider closes down/cuts off free access.

jancborchardt:
@v1r0x ok, let's go for Mapzen then. :)

Btw I think it would be good to merge this as early as possible so the PR doesn't become too big and people can also add improvements on top of the visible master. :)

jancborchardt:
(As current master does not work at all anyway.)

v1r0x:
> @v1r0x Mapbox is also used by GNOME Maps since Mapquest shut down their free access. An API key shouldn't be a huge deal, the weather app also requires one to work (pulling data from Openweathermap).

That's right. But should we add one (hardcoded) for all users or let them create one?

jancborchardt:
Let's say we add a default one for everyone for now. As soon as there's issues we can talk about customization.

v1r0x:
One question regarding the API key. Who should create the API Key? And are there any risks? E.g. abuse.

e-alfred:
You could do it like in the Weather app that moved this setting into the admin section recently. I do not think that a normal user should be bothered with such a setting after all, only the admin should set it onece and everyone can use an app then.

jancborchardt:
Yeah, API key should be an admin setting only. And would there be a default (or fair use) to ensure it works on simple install?

In any case, let's merge this branch as soon as possible @v1r0x, ok? :)

v1r0x:
I'm unsure about a private API key as default. I also already have a private mapzen key, which I need for some private projects. What do you think about a "official" nextcloud key which is the default? Is this even possible? @jancborchardt

jancborchardt:
How much could be used under that plan? Alternatively we could mail them about it and ask? cc @jospoortvliet

Also cc @freenerd who works at Mapbox, and @zeenix @mlundblad who work on GNOME Maps. How did you do it with the API key, and is there any way we could do it similarly with Nextcloud Maps?

Because we want to integrate Maps into Calendar, Contacts etc as well, so it makes sense to have a proper solution which works out of the box and doesn't need a technical person get an API key. :)

v1r0x:
> How much could be used under that plan?

screenshot-2017-10-21 mapzen flex mapzen

jancborchardt:
@v1r0x mind emailing both Mapzen and Mapbox to inquire if we can partner with them in some way?

@freenerd @zeenix @mlundblad any input how you handle it with GNOME Maps is appreciated :)

v1r0x:
@jancborchardt No problem. I'll cc you in the mail and attach this issue. Ok?

jancborchardt:
@v1r0x yep, great! :)

e-alfred:
I built the app and noticed a few thing that could be improved:

- Search should show a drop down menu with matching items (if multiple results are found)
- If geolocation is used a popover should be displayed for the location identical to an item found using the search
- As you use leaflet now, selecting different types of maps (satellite, terrain, routes for example) would be great (like in GpxPod)

e-alfred:
Game over for Mapzen, it will close down at the end of January:

https://mapzen.com/blog/migration/

I hope not a lot of work was done until now because it would be wasted once again. This also shows that the dependency on external services is extremely bad after all. Striving for self reliance or at least some kind of fallback sould be seriously considered, I think.

v1r0x:
I tried to talk to mapzen devs, but didn't work out. So, not much work was done.
The leaflet-routing-machine is used (afair) in the old version and had a look at it for the rework, but afair it was very limited with plain osm. Thus, I decided to switch to something else. In general osm is very limited.
Maybe osrm is worth looking at it.

e-alfred:
Seems like @eneiluj is now working on this app! @eneiluj Do you have some release we can test out? That would be awesome!

eneiluj:
@e-alfred You can try rework-eneiluj branch. It's functional. I've written what's been done so far in #32. If we all agree, we (@tacruc and me) could work on rework branch and continue discussing in this PR thread instead of #32.

jancborchardt:
@eneiluj awesome! Could you open a pull request based on that branch, then we can use that to discuss and collaborate. :)

eneiluj:
@jancborchardt Should I make a PR to `master` or to `rework` ?

jancborchardt:
@eneiluj since it’s based on rework but rework has been inactive, I think making the pull request to master is best. (cc @v1r0x just FYI)

eneiluj:
@jancborchardt Here comes #44. Maps app is back on track :wink:.

PR #40, opened 2019-04-06
Fixes that with two contacts one is allways visible even if it should…

tacruc:
… not


eneiluj:
Yep, merged.

PR #41, opened 2019-04-09
Tacruc place photos

tacruc:
Updating photo position based on drag and drop. At the moment the first positon shown is the current position of the user. If  coordinates are provided by the backend this coordinates are used.

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -119,12 +119,12 @@
             <field>
                 <name>lat</name>
                 <type>float</type>
-                <notnull>true</notnull>
+                <notnull>false</notnull>
             </field>
             <field>
                 <name>lng</name>
                 <type>float</type>
-                <notnull>true</notnull>
+                <notnull>false</notnull>
             </field>
             <field>
                 <name>date_taken</name>
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -18,6 +18,7 @@ return [
 
         // photos
         ['name' => 'photos#getPhotosFromDb', 'url' => '/photos', 'verb' => 'GET'],
+        ['name' => 'photos#getNonLocalizedPhotosFromDb', 'url' => '/photos/nonlocalized', 'verb' => 'GET'],
         ['name' => 'photos#placePhotos', 'url' => '/photos', 'verb' => 'POST'],
 
         // contacts
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
     "require": {
-        "lsolesen/pel": "^0.9.6"
+        "lsolesen/pel": "^0.9.6",
     }
 }
--- a/css/style.css
+++ b/css/style.css
@@ -308,10 +308,8 @@ tr.selected td {
 }
 
 .leaflet-marker-nonLocalizedPhoto {
-    border: 2px solid #fff;
+    border: 2px solid #ff0000;
     box-shadow: 3px 3px 10px #888;
-    -webkit-filter: grayscale(100%); /* Safari 6.0 - 9.0 */
-    filter: grayscale(100%);
 }
 
 .leaflet-marker-nonLocalizedPhoto.nonLocalizedPhoto-marker{
@@ -324,7 +322,7 @@ tr.selected td {
     bottom: 16px;
     border-width: 10px 10px 0;
     border-style: solid;
-    border-color: #fff transparent;
+    border-color: #ff0000 transparent;
     display: block;
     width: 0;
     margin-left: auto;
@@ -338,6 +336,8 @@ tr.selected td {
     background-position: center center;
     background-repeat: no-repeat;
     background-color: white;
+    -webkit-filter: grayscale(100%); /* Safari 6.0 - 9.0 */
+    filter: grayscale(100%);
 }
 
 .leaflet-marker-nonLocalizedPhoto .label {
@@ -417,6 +417,7 @@ table.editFavorite input[type=text] {
 }
 #toggleRoutingButton button,
 #togglePhotosButton button,
+#toggleNonLocalizedPhotosButton button,
 #toggleContactsButton button,
 #toggleFavoritesButton button,
 #toggleTracksButton button,
@@ -427,6 +428,7 @@ table.editFavorite input[type=text] {
 #navigation-routing > a,
 #navigation-favorites > a,
 #navigation-photos > a,
+#navigation-nonLocalizedPhotos > a,
 #navigation-contacts > a,
 #navigation-tracks > a,
 .track-line > a,
--- a/css/style.css
+++ b/css/style.css
@@ -307,6 +307,57 @@ tr.selected td {
     opacity: 1 !important;
 }
 
+.leaflet-marker-nonLocalizedPhoto {
+    border: 2px solid #fff;
+    box-shadow: 3px 3px 10px #888;
+    -webkit-filter: grayscale(100%); /* Safari 6.0 - 9.0 */
+    filter: grayscale(100%);
+}
+
+.leaflet-marker-nonLocalizedPhoto.nonLocalizedPhoto-marker{
+    top: -10px;
+}
+
+.leaflet-marker-nonLocalizedPhoto.nonLocalizedPhoto-marker:after {
+    content:"";
+    position: relative;
+    bottom: 16px;
+    border-width: 10px 10px 0;
+    border-style: solid;
+    border-color: #fff transparent;
+    display: block;
+    width: 0;
+    margin-left: auto;
+    margin-right: auto;
+}
+
+.leaflet-marker-nonLocalizedPhoto .thumbnail {
+    width: 100%;
+    height: 100%;
+    background-size: cover;
+    background-position: center center;
+    background-repeat: no-repeat;
+    background-color: white;
+}
+
+.leaflet-marker-nonLocalizedPhoto .label {
+    position: absolute;
+    top: -7px;
+    right: -11px;
+    color: #fff;
+    background-color: #333;
+    border-radius: 9px;
+    height: 18px;
+    min-width: 18px;
+    line-height: 12px;
+    text-align: center;
+    padding: 3px;
+}
+
+.leaflet-marker-nonLocalizedPhoto-tooltip {
+    opacity: 1 !important;
+}
+
 .easy-button-button {
     margin: 0 !important;
     padding: 0 !important;
--- a/js/nonLocalizedPhotosController.js
+++ b/js/nonLocalizedPhotosController.js
@@ -285,9 +285,24 @@ NonLocalizedPhotosController.prototype = {
     saveCordinatesToImage : function (marker) {
         var latlng = marker.getLatLng();
         this.photosController.placePhotos([marker.data.path], latlng.lat, latlng.lng);
-        this.nonLocalizedPhotoLayer.removeLayer(marker);
-        this.nonLocalizedPhotoMarkers.pop(marker);
-        this.nonLocalizedPhotoMarkersLastVisible = this.nonLocalizedPhotoMarkersLastVisible - 1;
+        var date = marker.data.date;
+        var path = marker.data.path;
+        var removedMarkers = [];
+        for (var i = this.nonLocalizedPhotoMarkersFirstVisible; i < this.nonLocalizedPhotoMarkers.length && this.nonLocalizedPhotoMarkers[i].data.date <= date; i++) {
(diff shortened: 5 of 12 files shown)


eneiluj:
Nice. I wonder how it's possible to handle the case when there are lots of non-localized photos. It's a bit weird to place them all at current position.

I guess this is temporary and you plan to make the guessing part solve that. But still, what about if there are many photos with no guess?

What is the way you want it to be at the end?

eneiluj:
BTW nice to make this happen in a new controller to avoid messing with photosController!

tacruc:
First there is still the time filter. Which means you cloud filter the photos oft your last trip etc.

So the remainig Problem would be a performance Problem, because the biggest possible set is loaded in the beginning. For that I would do it as in the photos controller and tackle it, if we know where it is (we know that it is above 500 photos)

eneiluj:
That's no problem to load a lot of photos, thumbnails are lazy-loaded and a big amount of markers is not a problem thanks to clustering.

> For that I would do it as in the photos controller and tackle it

What do you mean?

> if we know where it is (we know that it is above 500 photos)

What do you mean?

tacruc:
I just mean that the performance problem is not a problem for now.

tacruc:
For the guessing part I will have to merge the track branch. So I'm waiting with it until it is merged.

eneiluj:
Tracks branch merge is coming soon. I could even merge it now as the only thing i want to add is tracks statistics table.

tacruc:
If you merge now, I would probably work on the guessing funktion this evening.

eneiluj:
I'll merge before this evening.

eneiluj:
Guessing is probably not going to be a piece of cake. As tracks data is not in the DB, i don't know if guessing can be done on the server side. We can discuss that later.

eneiluj:
Device point will be stored in DB.

For the base feature, it's useless to store tracks content in DB. Only the file ID is stored in DB. In the guessing part, server side, you could still load all tracks content (only those in maps DB) from corresponding files as there won't be many of them.

tacruc:
I should just do it in a clever way. To not load all files for each photo.

eneiluj:
Yes, I meant loading it once before guessing for all pictures of course.

tacruc:
Do you know why the navigation entry is not colored like the other entries and changing activity if the mouse is over it?

eneiluj:
I forced the opacity in style.css.

eneiluj:
Let's not worry about the conflict in `PhotofilesService.php`. Easy to fix.

Feel free to tell me when this PR needs some testing or feedback.

tacruc:
I think you cloud test it, I cloud not do to intensitiv test, do to lack of gpx files.

eneiluj:
I get the overall meaning of the sentence. I guess you mean:

"I think you could test it, I could not do very intensive tests, due to a lack of gpx files." :wink:

The feature is nice! It would require some hints in the UI to let the user understand what happens and when and why.

I guess it's not definitive but one problem of current state: guessing is done thanks to loaded tracks at the time the photos are added to DB. This means if there is no track loaded when photos are added by file hooks, there is nothing to guess and guessing is not done again later. This would need a little brainstorm. Maybe we could let the user trigger the guessing process manually so that he/she can first load a few tracks. If the manual guessing is asked, we could set the slider values to only see photos with a guess. Anyway we can think about that later.

Here are my remarks after having red the code:

* could you use spaces instead of tabs? There's a mix of both.
* could you comment a little bit more? especially methods in `GeophotoService.php` for which it's not obvious what they do.
* could you separate photo placement with single latlng and with multiple latlngs? I prefer a little bit of code duplication than having ambiguous methods :smile:. I mean: could you write a new method to place multiple photos with multiple latlngs instead of modifying `setFilesCoords()` and making tests like `is_array`?  It could be named `setFilesCoordsMultiple($userId, $paths, $lats, $lngs)` with 's' to express that there are multiple coordinates. It would be called by `setPhotosFilesCoordsMultiple` in the same controller which would be called by `placePhotosMultiple` in `PhotosController.php` which would be the target of a new route (`['name' => 'photos#placePhotosMultiple', 'url' => '/photos-multiple', 'verb' => 'POST'],` for example). You can still call `this.photosController.placePhotos()` in `saveCordinatesToImage` but you would need to implement a new method (`placePhotosMultiple`) called in `menuSaveAllVisible`. I guess this `placePhotosMultiple` method fits better in `photosController.js` than in `nonLocalizedPhotosController.js`. I hope i'm clear enough. Sorry about the excessive details. I can do it if you don't have enough time or you don't feel like it. Just tell me.

I'll make some tests soon.

eneiluj:
I edited previous comment which was wrong about the guessing problem.

eneiluj:
Also, I don't know if it's useful to display all non-localized pictures on the map. I think only getting the ones with a guess is enough.

When all non-localized photos are displayed at the same position, it's strange to have so many of them. Even if the user filters to one day, (the day of a hike for example) there could still be dozens of photos at the same position so it requires to spiderfy the big cluster to be able to drag'n'drop them to place them... Not very handy.

tacruc:
> I get the overall meaning of the sentence. I guess you mean:
> 
> "I think you could test it, I could not do very intensive tests, due to a lack of gpx files." wink
> 

Yes

> The feature is nice! It would require some hints in the UI to let the user understand what happens and when and why.
> 

If it is finished I i have time I would look into an on screen manual. Saying click a photo to save the coordinates, And drag the photo to change them.

Maybe explain the change after saving. And add a right click option on the photo to delete the coordinates. So that the photo is than handled by the nonlocalziedphotoControler. I would add a red boarder around the nonlocalziedPhotos because the black/white stuff is not that visible.

> I guess it's not definitive but one problem of current state: guessing is done on each page load. This would need a little brainstorm. Maybe we could let the user trigger the guessing process manually so that he/she can first load a few tracks. If the manual guessing is asked, we could set the slider values to only see photos with a guess. There would be multiple guessing type (tracks, devices). Anyway we can think about that later.
> 

It's not done on pageload, but on click of the show nonlocalized Pictures. 
What is your problem here, you want to address?

- If it is speed or loading time:

Yes it is slow at the moment, took 22s for 900 pictures and 50 tracks on my pc. I would not add more UI stuff, but make the backend smarter. I have to check what is the easiest here. We cloud load the pictures incrementally. Loading all picture would still take the same time, but at least the is immediately something happening.  To make it faster one would need some caching in the backend.

- If it is usability. I clicked the button and than uploaded a new track:

I think we should in general make something that hiding and showing elements checks for changes in the backend.


> Here are my remarks after having red the code:
> 
>     * could you use spaces instead of tabs? There's a mix of both.
> 

Yes, sorry so I have to check my editor again.

>
>     * could you comment a little bit more? especially methods in `GeophotoService.php` for which it's not obvious what they do.
> 

Will do that this evening.

>
>     * could you separate photo placement with single latlng and with multiple latlngs? I prefer a little bit of code duplication than having ambiguous methods smile. I mean: could you write a new method to place multiple photos with multiple latlngs instead of modifying `setFilesCoords()` and making tests like `is_array`?  It could be named `setFilesCoordsMultiple($userId, $paths, $lats, $lngs)` with 's' to express that there are multiple coordinates. It would be called by `setPhotosFilesCoordsMultiple` in the same controller which would be called by `placePhotosMultiple` in `PhotosController.php` which would be the target of a new route (`['name' => 'photos#placePhotosMultiple', 'url' => '/photos-multiple', 'verb' => 'POST'],` for example). You can still call `this.photosController.placePhotos()` in `saveCordinatesToImage` but you would need to implement a new method (`placePhotosMultiple`) called in `menuSaveAllVisible`. I guess this `placePhotosMultiple` method fits better in `photosController.js` than in `nonLocalizedPhotosController.js`. I hope i'm clear enough. Sorry about the excessive details. I can do it if you don't have enough time or you don't feel like it. Just tell me.
> 

I totally agree with you, normally I use the "s" in the end to tell that it is multiple. So I expected setFilesCoords to set coordinates for multiple files. But the s on Coordinates was somehow just refering to lat lng being allready multiple Values. 
So I thought the nameing would be like:
1. Single file single Lat, Lng
`setFileCord`
2. Multiple Files single (Lat,Lng)
`setFilesCord`
3. Single file multiple (lat, lng)
`setFileCords`
4. Multiple Files multiple (lat,lng)
`setFilesCords`
So I just included the check array to not break your code.

So what your suggestion is:
1. Single file single Lat, Lng
`not interessting`
2. Multiple Files single (Lat,Lng)
`setFilesCords`
3. Single file multiple (lat, lng)
`not interessting`
4. Multiple Files multiple (lat,lng)
`setFilesCordsMultiple`

What do you think of:
1. Single file single Lat, Lng
`not interessting`
2. Multiple Files single (Lat,Lng)
`setFilesLatLng`
3. Single file multiple (lat, lng)
`not interessting`
4. Multiple Files multiple (lat,lng)
`setFilesLatsLngs`

Somehow I'm not happy with all the suggestions. But at the moment I can not come up with something good. So feel free to suggest or implement what ever you like, and I will be happy with it.

> 
> I'll make some tests soon.

tacruc:
> Also, I don't know if it's useful to display all non-localized pictures on the map. I think only getting the ones with a guess is enough.
> 
> When all non-localized photos are displayed at the same position, it's strange to have so many of them. Even if the user filters to one day, (the day of a hike for example) there could still be dozens of photos at the same position so it requires to spiderfy the big cluster to be able to drag'n'drop them to place them... Not very handy.

I agree, we could show them if it is less than x. Where we have to find a good value for x. As a starting point I would say 10.

tacruc:
The is an issue that all pictures are shown if a picture is placed and the photo layer was not visible before.

eneiluj:
Ok i agree my suggestion is not very nice. I guess it's much simpler and clearer to always give lats and lngs arrays to setFilesCoords. The JS function would be `placePhotos: function(paths, lats, lngs, directory=false)`. Let's keep just one route and remove the `is_array` test. The test becomes : if the lats and lngs array's length is `> $i` we get it at `$i`, else we get the first element. This way there is no type test.

Better? If you're ok, i'll do it when merging.

> The is an issue that all pictures are shown if a picture is placed and the photo layer was not visible before.

I'll solve that after we make a first merge to rework-again.

> It's not done on pageload, but on click of the show nonlocalized Pictures.
What is your problem here, you want to address?

Yep right, it's done when the data is loaded. My concerns are both performance and usability :wink:. How is the user supposed to know he/she has to add tracks before loading non-localized photos to get correct guess? 

An optimization to reduce the number of photos processed by the guessing part would be to first get the min and max dates of tracks and then query photos in the DB in this range.

> I agree, we could show them if it is less than x. Where we have to find a good value for x. As a starting point I would say 10.

We can think about that later.

tacruc:
> Ok i agree my suggestion is not very nice. I guess it's much simpler and clearer to always give lats and lngs arrays to setFilesCoords. The JS function would be `placePhotos: function(paths, lats, lngs, directory=false)`. Let's keep just one route and remove the `is_array` test. The test becomes : if the lats and lngs array's length is `> $i` we get it at `$i`, else we get the first element. This way there is no type test.
> 
> Better? If you're ok, i'll do it when merging.
> 
> > The is an issue that all pictures are shown if a picture is placed and the photo layer was not visible before.
> 
> I'll solve that after we make a first merge to rework-again.
> 
> > It's not done on pageload, but on click of the show nonlocalized Pictures.
> > What is your problem here, you want to address?
> 
> Yep right, it's done when the data is loaded. My concerns are both performance and usability wink. How is the user supposed to know he/she has to add tracks before loading non-localized photos to get correct guess?
> 

Personally I think that the bigger use case is actually the devices part. But anyway a hint or a tooltip on first activation telling that the geolocation is callculated from the tracks and the devices is usefull. I would not force a user to select tracks or devices before something is done. 

> An optimization to reduce the number of photos processed by the guessing part would be to first get the min and max dates of tracks and then query photos in the DB in this range.
> 

There is plenty of space for optimization. Just a list of ideas
- Cache: Save the guesses, update on photo or gpx change
- incidental loading on the request
- do not go though the full track but check first and last timestamp first
- do not go element by element in the track, but use a logarithmic search 
- Safe the tracks in some sorted struckture to use a sorted search for tracks with the right range
- only select pictures, for which exists tracks
- Do guessing in the background and not on request.
- make parallel request for pictures and share the timeLocationSetData between requests 

But lets keep in mind that even with the not optimized code it is not too slow, and maybe a small hind, with "processing your request, this cloud take a while, please wait", would be enough for now.

> > I agree, we could show them if it is less than x. Where we have to find a good value for x. As a starting point I would say 10.
> 
> We can think about that later.

PR #42, opened 2019-04-13
register​Protocol​Handler() for geo: url

tacruc:


eneiluj:
Nice!
I've made it safer : don't try to get geolink coordinates in main page (index).
I've added a right click action to generate a geo link and copy it to clipboard.

Issue #45, opened 2019-04-23
ToDo non localized pictures


tacruc:
### Backend

- [ ] Different gpx loading to allow for big gpx files as coming from google timeline.
- [ ] do the matching smarter.
- [ ] Cache TimeordedPointSets for x seconds
- [ ] Change API for lazy loading
- [x] Read Point from gpx wptType, rteType, ~~trkType which are not in a trkseg (allready done)~~

### Frontend
- [ ] lazy load nonlocalizedPhots
- [ ] restructure  Photo menu entry, plug bot photos in one main menu entry

eneiluj:
> lazy load nonlocalizedPhotos

What do you mean by that? They already are lazy loaded when needed. Could you more precise?

About photos, @jancborchardt said there shouldn't be any difference between localized and non-localized photos in terms of interface. I don't have a strong opinion about that. I would say it's nice to at least make the popup have a different color (like warning color currently used).

About the menu entries, I think I agree with @jancborchardt it would be nicer to have just one named Photos which would show localized AND non-localized photos.

tacruc:
> > lazy load nonlocalizedPhotos
> 
> What do you mean by that? They already are lazy loaded when needed. Could you more precise?

I mean the loading of the marker location, not of the image. 

1. A request where the list of nonlocalized Id's is loaded.
2. One request resolving the set of (lat, lng) pair for each photo.


```

About photos, @jancborchardt said there shouldn't be any difference between localized and non-localized photos in terms of interface. I don't have a strong opinion about that. I would say it's nice to at least make the popup have a different color (like warning color currently used).

About the menu entries, I think I agree with @jancborchardt it would be nicer to have just one named Photos which would show localized AND non-localized photos.
```

I think this is similar to option A, because the sub entries would be toggled as default anyways. I don't have a strong opinion if they should exist or not. But if it doesn't exists we shouldn't show the pictures without guess at the current position, or what do you thin?

jancborchardt:
> About photos, @jancborchardt said there shouldn't be any difference between localized and non-localized photos in terms of interface. I don't have a strong opinion about that. I would say it's nice to at least make the popup have a different color (like warning color currently used).

What we could do is in the second row (where the filename is shown) just put (estimated location) in parentheses like so:

> **Tuesday, 19 July 2017, 20:34**
> IMG_09448.jpg (estimated location)

What do you think?


[...]

PR #46, opened 2019-04-23
Non localized pictures devices

tacruc:
Addressing toDo in #45

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -21,6 +21,8 @@ return [
         // photos
         ['name' => 'photos#getPhotosFromDb', 'url' => '/photos', 'verb' => 'GET'],
         ['name' => 'photos#getNonLocalizedPhotosFromDb', 'url' => '/photos/nonlocalized', 'verb' => 'GET'],
+        ['name' => 'photos#getNonLocalizedPhotoIds', 'url' => '/photos/nonlocalizedids', 'verb' => 'GET'],
+        ['name' => 'photos#getNonLocalizedPhotosById', 'url' => '/photos/nonlocalized/{id}', 'verb' => 'GET'],
         ['name' => 'photos#placePhotos', 'url' => '/photos', 'verb' => 'POST'],
 
         // contacts
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -22,7 +22,7 @@ return [
         ['name' => 'photos#getPhotosFromDb', 'url' => '/photos', 'verb' => 'GET'],
         ['name' => 'photos#getNonLocalizedPhotosFromDb', 'url' => '/photos/nonlocalized', 'verb' => 'GET'],
         ['name' => 'photos#getNonLocalizedPhotoIds', 'url' => '/photos/nonlocalizedids', 'verb' => 'GET'],
-        ['name' => 'photos#getNonLocalizedPhotosById', 'url' => '/photos/nonlocalized/{id}', 'verb' => 'GET'],
+        ['name' => 'photos#getNonLocalizedPhotosByIds', 'url' => '/photos/nonlocalizedbyids', 'verb' => 'GET'],
         ['name' => 'photos#placePhotos', 'url' => '/photos', 'verb' => 'POST'],
 
         // contacts
--- a/js/nonLocalizedPhotosController.js
+++ b/js/nonLocalizedPhotosController.js
@@ -14,6 +14,7 @@ function NonLocalizedPhotosController (optionsController, timeFilterController, 
     this.nonLocalizedPhotoMarkersLastVisible = -1;
     this.timeFilterBegin = 0;
     this.timeFilterEnd = Date.now();
+    this.nonLocalizedPhotoIds = [];
 }
 
 NonLocalizedPhotosController.prototype = {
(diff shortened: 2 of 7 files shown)


tacruc:
For the lazy loading ICache is used. There is the issue that the cached entry is not deleted https://github.com/nextcloud/server/issues/15208. 

@eneiluj Should we programm a workaround and safe the timestamp manually in the cache and programm a deletion, or should we wait unit the issue is fixed on the server side?

eneiluj:
Yay, seems nice!

It would be great if you could explain what you did in a few lines. A list of things you did technically, why you did it and what effect/benefit it has.

> Should we programm a workaround and safe the timestamp manually...

I don't know since I didn't really get what the issue is. Even if the cache is not deleted, it probably does not generate a huge amount of information. I think we can implement Maps like if caching was working as expected.

What do you mean by "save the timestamp manually"? Do you mean without caching?

A first remark: `callForImages` was getting all nonLocphotos information in one request before, why did you make it do a request per photo? I'm sure there is a reason but I can't find out which :smile:.

My point of view about getting a lot of elements with ajax requests is: by experience it's much faster and efficient (client CPU load) to get them all in one request if you want them all anyway. It's faster because making a HTTP request costs extra time which is probably bigger than the time needed to process the result. So if you repeat it hundreds of times, the time proportion taken by HTTP requests is huge. It's more efficient because you can factorize the processing of results.

I'm sorry if I say potentially obvious things or if I didn't exactly get the reasons of your choices :smile:. I just want to be clear and explicit :nerd_face:.

Ow and a very small remark: You fixed `$this->timeordedPointSets` by naming it `$this->timeorderedPointSets` but you didn't fix the loop variable name : `foreach ($this->timeorderedPointSets as $timeordedPointSet) {`

tacruc:
> Yay, seems nice!
> 
> It would be great if you could explain what you did in a few lines. A list of things you did technically, why you did it and what effect/benefit it has.
> 
> > Should we programm a workaround and safe the timestamp manually...
> 
> I don't know since I didn't really get what the issue is. Even if the cache is not deleted, it probably does not generate a huge amount of information. I think we can implement Maps like if caching was working as expected.
> 
> What do you mean by "save the timestamp manually"? Do you mean without caching?
> 
I will implement it, I was misunderstanding the ICache documentation.

> A first remark: `callForImages` was getting all nonLocphotos information in one request before, why did you make it do a request per photo? I'm sure there is a reason but I can't find out which smile.
> 

The problem was that the lookup of the location can take some time, it scales linear in the number of tracks + number of devices and number of pictures. I had that the big request took 10 seconds. So you would click on the show pictures and it took 10 seconds until something was shown. Now the first picture is shown after 500ms. Which means that the time the user is waiting without visual feedback is reduced. On the other hand the looking up of the location is done by multiple threads now. 
On the other hand now the reading of the gpx files and devices files and the preprocessing is done on pageload so in advanced. 

> My point of view about getting a lot of elements with ajax requests is: by experience it's much faster and efficient (client CPU load) to get them all in one request if you want them all anyway. It's faster because making a HTTP request costs extra time which is probably bigger than the time needed to process the result. So if you repeat it hundreds of times, the time proportion taken by HTTP requests is huge. It's more efficient because you can factorize the processing of results.
> 
> I'm sorry if I say potentially obvious things or if I didn't exactly get the reasons of your choices smile. I just want to be clear and explicit nerd_face.
> 
> Ow and a very small remark: You fixed `$this->timeordedPointSets` by naming it `$this->timeorderedPointSets` but you didn't fix the loop variable name : `foreach ($this->timeorderedPointSets as $timeordedPointSet) {`

I will fix that

eneiluj:
Ok I think I got it now. Neat!

Without explanation and without reading the code in detail it was hard for me to understand.

So from what I understand: Now, when Photos layer is enabled, js controller first gets the ids of all non-localized photos and then does one request per photo to ask server side controller to guess the coordinates and provide the photo entry. Each successful guess is cached on the server side.

This PR is not about pre-guessing with a cron-job or files/contacts/anything hooks.

Am I right?

So I agree with you about making multiple requests because the guessing is done at this moment. I thought you were just getting already guessed things from the DB.

> On the other hand now the reading of the gpx files and devices files and the preprocessing is done on pageload so in advanced.

Is it really done on page load or just when Photos layer is getting enabled?

We need to be very explicit and descriptive when discussing those kind of features otherwise I say stupid things. :smile:

eneiluj:
Btw,

> I was misunderstanding the ICache documentation

Does it mean there is no bug in NC caching system?

tacruc:
Yes the ttl is just the minimum time to life you should not expected it to be deleted after it

tacruc:
The main Idea was not to wait for 10000 pictures that the location is computed, but to load them incrementally. I think loading every picture was too much, so I will change it to load batches of 20 pictures or something.

tacruc:
Now it is doing what it is supposed to do. Reducing the loading of pictures without geoinformation.
On pageload a list of all pictures is loaded. With this request the gpx files are loaded in the backend. Then it is fast to load batches of 250 Photos.
Reduces the loading time on my machine with 409 pictures from 4s to 0.9 seconds.

tacruc:
I will close this

PR #48, opened 2019-04-25
Tacruc non localized pictures devices new

tacruc:
Loading positions from devices, too.
Removed the lazy loading.


tacruc:
At the current state i would prefere differently loading, because the loading is a way slower. 
If the loading is speeded up, where I still have open questions, I'm fine with loading and showing them together.

PR #53, opened 2019-05-01
another bunch of commits

eneiluj:
Maps is now almost @jancborchardt -compliant :wink:.
In short:
* a few design fixes
* a lot of js fixes
* a few string fixes (shorter)
* error message when gpx track parsing fails
* use DB Migration system
* now able to search POIs like restaurant, bar etc...inside current map view (multiple results displayed)
* import device(s) from gpx/kml/kmz
* export device(s) to gpx
* export single favorite category
* Files default action for gpx files: open in Maps as a track
* right click on anything in side bar: open corresponding menu

CI tests didn't change (still pretty empty) but it fails because of a recent change in Nextcloud which makes it fail to lazy-create user folders after creation.


jancborchardt:
Gooood stuff! :) Am traveling at the moment so can’t test so much, but will hopefully have time to do another review session end of the week.

cc @nextcloud/maps folks for review :tada:

tacruc:
The right click in the menu doesn't work for me the menu is opened onclick and closed mouse release

eneiluj:
@tacruc Thanks for your feedback.

> The right click in the menu doesn't work for me the menu is opened onclick and closed mouse release

Tell me more about your context so I can reproduce the problem. Browser, browser version, OS and anything that can help is welcome.

> If I open a track from the files app the hight is overlaying with he time slider.

Do you mean it's different from classic elevation display when asked in "normal" page? I guess not. So the problem is related to elevation control in any case. Well we could add a bottom margin but with high resolutions or strange aspect ratio there would still be a potential overlapping.

In the future, it's probably better if we don't push in the same branch so that we can't review our own code :wink:.

tacruc:
> @tacruc Thanks for your feedback.
> 
> > The right click in the menu doesn't work for me the menu is opened onclick and closed mouse release
> 
> Tell me more about your context so I can reproduce the problem. Browser, browser version, OS and anything that can help is welcome.
> 
Firefox on Opensuse.
The callback in script.js line 46-53 is executed two. Line 52 removes the menu.
```
        window.onclick = function(event) {
            $('.leaflet-control-layers').hide();
            $('.easy-button-container').show();
            if (!event.target.matches('.app-navigation-entry-utils-menu-button button')) {
                $('.app-navigation-entry-menu.open').removeClass('open');
            }
            mapController.map.contextmenu.hide();
        };
```
> > If I open a track from the files app the hight is overlaying with he time slider.
> 
> Do you mean it's different from classic elevation display when asked in "normal" page? I guess not. So the problem is related to elevation control in any case. Well we could add a bottom margin but with high resolutions or strange aspect ratio there would still be a potential overlapping.
> 
> In the future, it's probably better if we don't push in the same branch so that we can't review our own code wink.
I agree it is not related here.

eneiluj:
Mmh I guess this is Firefox 66... I'm using Firefox 68 nightly and it strangely works fine. Anyway there was a mistake that is fixed now.

For the elevation-slider overlay, I'll think about that later. Do you want to wait until we figure this out to merge?

PR #55, opened 2019-05-02
Tacruc pel new

### LINKED ISSUE #47: exif_read_data is making problems
On most of my jpegs php read exif is throwing errors. This is giving me a hard time to get 1000 picture in my test installation to do some performance testing. 

`exif_read_data(P1050526.JPG): Process tag(x013B=Artist ): Illegal components(0) at /srv/www/htdocs/server/lib/private/legacy/image.php#418`

I didn't manage to make it more stable, but the metadata app is reading them correctly. 
Some sample picture:
P1010498

tacruc:
fixes #47 is depending on #53

### DIFF
--- a/lib/Service/GeophotoService.php
+++ b/lib/Service/GeophotoService.php
@@ -35,7 +35,7 @@ class GeophotoService {
     private $logger;
     private $preview;
     private $tracksService;
-    private $timeordedPointSets;
+    private $timeorderedPointSets;
     private $devicesService;
 
     public function __construct (ILogger $logger, IRootFolder $root, IL10N $l10n, GeophotoMapper $photoMapper, IPreview $preview, TracksService $tracksService, DevicesService $devicesService, $userId) {
@@ -45,7 +45,7 @@ class GeophotoService {
         $this->logger = $logger;
         $this->preview = $preview;
         $this->tracksService = $tracksService;
-        $this->timeordedPointSets = null;
+        $this->timeorderedPointSets = null;
         $this->userId = $userId;
         $this->devicesService = $devicesService;
 
@@ -94,7 +94,7 @@ class GeophotoService {
      * @return array with geodatas of all nonLocalizedPhotos
      */
     public function getNonLocalizedFromDB ($userId) {
-        $foo = $this->loadTimeordedPointSets($userId);
+        $foo = $this->loadTimeorderedPointSets($userId);
         $photoEntities = $this->photoMapper->findAllNonLocalized($userId);
         $userFolder = $this->getFolderForUser($userId);
         $filesById = [];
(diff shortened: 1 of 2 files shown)


eneiluj:
Cool. Thanks. Sounds good. I'll try it soon and merge.

v1r0x:
I'd recommend pel as default and use `exif_read_data` as backup. Last time I had to decide which to use, pel had more features and also supports maker notes (= exif properties defined by camera maker; at least supported for canon cameras).

eneiluj:
@v1r0x I agree with using `pel` by default. @tacruc What do you think about that?

@tacruc Tell me when you feel it's ready so that I can make some tests and merge.

tacruc:
> I'd recommend pel as default and use `exif_read_data` as backup. Last time I had to decide which to use, pel had more features and also supports maker notes (= exif properties defined by camera maker; at least supported for canon cameras).

I would agree in general, but the way it's done now works and we only need 3 Values which are supported by both libraries. Futhermore I had some memory leaks with Pel, thats why the exeptions have to be cleared. We can change it any times if we have a reason.

eneiluj:
Hey, `getExifPelBackup` sometimes fails because `$pelIfdExif` can be null (on line 444 of `lib/Service/PhotofilesService.php`). It could be tested just before: `if (is_null($pelExif) or is_null($pelIfdExif)) {` on line 441.

But there still is a problem: exif info is not found for some photos (same ones that produce a null `$pelIfdExif`). Here is one: [nc.jpg.txt](https://github.com/nextcloud/maps/files/3143776/nc.jpg.txt). 
`exif_read_data` manages to get coordinates from those photos.

tacruc:
> Hey, `getExifPelBackup` sometimes fails because `$pelIfdExif` can be null (on line 444 of `lib/Service/PhotofilesService.php`). It could be tested just before: `if (is_null($pelExif) or is_null($pelIfdExif)) {` on line 441.
> 
Thanks I will fix it
> But there still is a problem: exif info is not found for some photos (same ones that produce a null `$pelIfdExif`). Here is one: [nc.jpg.txt](https://github.com/nextcloud/maps/files/3143776/nc.jpg.txt).
> `exif_read_data` manages to get coordinates from those photos.

Is it a problem of PEL or how I used PEL? I guess, if it is a problem of PEL its would support the current state that PEL is the backup of exif_read_data.

eneiluj:
Sorry for the delay.

> Is it a problem of PEL or how I used PEL?

I don't know.

So, what do you all think about accepting this PR?

I'm ok with it: call `exif_read_data` first and if it returns nothing, use `Pel` as a backup solution.

As `Pel` (or @tacruc 's implementation) sometimes fails getting coordinates, let's hope there are just a few files that contain exif coordinates and that will cause trouble to both `exif_read_data` and `Pel`...

tacruc:
> Sorry for the delay.
No problem
> 
> > Is it a problem of PEL or how I used PEL?
> 
> I don't know.
> 
> So, what do you all think about accepting this PR?
:+1: 
> 
> I'm ok with it: call `exif_read_data` first and if it returns nothing, use `Pel` as a backup solution.
> 
> As `Pel` (or @tacruc 's implementation) sometimes fails getting coordinates, let's hope there are just a few files that contain exif coordinates and that will cause trouble to both `exif_read_data` and `Pel`...
At least they seam to be uncorrelated to the probability that one picture fails both seams to be small.

Issue #34, opened 2019-04-03
Display Contacts on Maps


tacruc:
### ToDo:

- [x] Caching Table for addresses to Cordinates (Address string), Lat, Lng, timelookedup
- [x] ~~When to look up the results, We should query less then one time per second, so one demand is not possible => should be done in the backend in advanced, to not hit the one request per second limitation~~
- [x] ~~Extend the navigation to show "suggestions for contacts"~~
 - [x] ~~Api call for contactSuggestions~~
- [x] phrase coordinat form vCard in backend and frontend
- [x] make contact preview nice
- [ ] Delete Geo attribute from vCard
- [ ] Set geo attibute according to vCard version, in vCard v4 it should be an geo url.
- [ ] Check permission when setting Geo Attribute

### Possible interesting vCard Attibutes

- ADR
  - Can be multiple, I guess all are interesting
- GEO
  - Different formates in v2,3 and v4

### Problems

* it's currently possible to place a contact (edit its coordinates) even if it's from an address book shared with me with readonly permissions.

eneiluj:
Yep, I think vCard GEO attribute should be used in priority if it's there. Then if there are multiple addresses, i guess one could be used arbitrarily.

There's more, considering we have an address, there might be multiple search result. We can reasonably assume it's ok to take the first result.

Photo metadata are generated by the file hook. How could we proceed for Contacts? Is there a hook when a contact is created/updated? Or should we access contact database on each Maps page load? It would be smart to make the address search just once and store the result coordinates in Maps DB :smile:.

A lot to think about.

tacruc:
Well why would it be bedder to have it in extra table? There are hooks on carddav. But I think I would just read it from the db or an api.

eneiluj:
So maps DB could act like a sort of cache: first time a contact is loaded from contact DB, data is generated and stored, next time it's just retrieved from DB.

Is there a way to know if a contact was modified since a precise date?

eneiluj:
> Well why would it be bedder to have it in extra table?

To store potential coordinates resulting from geocoder search and avoid calling nominatim each time we load the contacts.

tacruc:
We cloud maybe use  
`public OCP\Contacts\IManager::search($pattern, $searchProperties=array(), $options=array())`

eneiluj:
Nice to see the new contact controller.

About the contact base64 picture: no problem, `<img src="BASE64_STRING"/>` works fine.

[...]

PR #57, opened 2019-05-16
Looking up contact addresses

tacruc:
This PR addresses #34 the part of the contacts address lookup.

We have the restriction that we're only allowed to lookup once a second on noamintim.  This is why the addresses are cached in the address_geo table.  

Lookup procedure to minimize request:

Address in db?
  - Yes
    - Looked up externally -> used cached $lat $lng
    - not Looked up externally -> try lookup Externally -> use that $lat lng
 - No
    - try to lookup externally -> add entry to database

On try lookup externally

- Last lookup older than 1 second -> lookup
- else -> return null, null

@eneiluj @jancborchardt any suggestions?

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -12,8 +12,9 @@
 namespace OCA\Maps\AppInfo;
 
 use OCP\AppFramework\App;
-
+use OCA\Maps\Service\AddressService;
 use OCP\Util;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 $app = new Application();
 $container = $app->getContainer();
@@ -23,6 +24,18 @@ $eventDispatcher->addListener('OCA\Files::loadAdditionalScripts', function() {
     Util::addScript('maps', 'filetypes');
     Util::addStyle('maps', 'filetypes');
 });
+
+// carddav/caldav lookup addresses
+$listener = function($event) use ($container) {
+    if ($event instanceof GenericEvent) {
+        $c = $event->getArgument('cardData');
+        $a = $container->query(AddressService::class);
+        $a->scheduleVCardForLookup($c);
+    }
+};
+
+$eventDispatcher->addListener('\OCA\DAV\CardDAV\CardDavBackend::createCard', $listener);
+$eventDispatcher->addListener('\OCA\DAV\CardDAV\CardDavBackend::updateCard', $listener);
 
 $l = \OC::$server->getL10N('maps');
 
--- a/appinfo/application.php
+++ b/appinfo/application.php
@@ -13,6 +13,7 @@ namespace OCA\Maps\AppInfo;
 
 
 use OC\AppFramework\Utility\SimpleContainer;
+use OCA\Maps\Service\AddressService;
 use \OCP\AppFramework\App;
 use OCA\Maps\Controller\PageController;
 use OCA\Maps\Controller\UtilsController;
@@ -190,5 +191,6 @@ class Application extends App {
             }
         );
     }
+
 
 }
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -8,7 +8,7 @@
     <licence>agpl</licence>
     <author mail="eneiluj@posteo.net">Julien Veyssier</author>
     <author mail="kontakt+github@arne.email">Arne Hamann</author>
-    <version>0.0.8</version>
+    <version>0.0.9</version>
     <namespace>Maps</namespace>
     <category>multimedia</category>
     <bugs>https://github.com/nextcloud/maps/issues</bugs>
--- a/lib/BackgroundJob/LookupMissingGeoJob.php
+++ b/lib/BackgroundJob/LookupMissingGeoJob.php
@@ -28,6 +28,8 @@ class LookupMissingGeoJob extends QueuedJob {
     /**
      * LookupMissingGeoJob constructor.
      *
+     * A QueuedJob to lookup missing geo information of addresses
+     *
      * @param AddressService $service
      * @param IJobList $jobList
      */
@@ -39,7 +41,9 @@ class LookupMissingGeoJob extends QueuedJob {
 
     public function run($arguments) {
         \OC::$server->getLogger()->debug("Cronjob maps address executed");
+        //lookup at most 200 addresses
         if (!$this->addressService->lookupMissingGeo(200)){
+            //if not all addresses where looked up successfully add a new job for next time
             $this->jobList->add(LookupMissingGeoJob::class,[]);
         }
     }
--- a/lib/BackgroundJob/LookupMissingGeoJob.php
+++ b/lib/BackgroundJob/LookupMissingGeoJob.php
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * Nextcloud - maps
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Arne Hamann
+ * @copyright Arne Hamann 2019
+ */
+
+namespace OCA\Maps\BackgroundJob;
+
+use \OCA\Maps\Service\AddressService;
+use \OCP\BackgroundJob\QueuedJob;
+use \OCP\BackgroundJob\IJobList;
+use \OCP\AppFramework\Utility\ITimeFactory;
+
+class LookupMissingGeoJob extends QueuedJob {
+
+    /** @var AddressService */
+    private $addressService;
+
+    /** @var AddressService */
+    private $jobList;
+
+    /**
+     * LookupMissingGeoJob constructor.
+     *
+     * @param AddressService $service
+     * @param IJobList $jobList
+     */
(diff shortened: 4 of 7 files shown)


tacruc:
Handling multiple Address is hard at the moment hence
`\OCA\DAV\CardDAV\AddressBookImpl::vCard2Array `
only returns one address. 
```
			// The following properties can be set multiple times
			} else if (in_array($property->name, ['CLOUD', 'EMAIL', 'IMPP', 'TEL', 'URL'])) {
```

jancborchardt:
Since the initial requests can take longer, we could prioritize somehow intelligently, if possible:
- Favorited contacts first (but Contacts does not support favorites yet, does it @nextcloud/contacts?)
- Contacts which are in lots of groups next, as they are likely important
- If we can somehow pre-parse to prioritize based on the current view – e.g. I’m currently looking at Kenya, then first load the contacts which have "Kenya" and neighboring countries in their addresses.
- If all else fails, prioritize contacts with pictures, and with more info, as they are more likely to be important to you.

tacruc:
@jancborchardt I like the idea, but I think it is a little bit overkill at the moment. 
Because the first lookup will be done on the creation of the vCard. 

- If you only create one contact there is a good change that the external lookup will directly work. (How big does an inctance have to be that there is one vCard address change every second? Or that there is a high change that an other user changed an address in the same second.)

- So if you upload more than one vCard at once the changes is hight that the external lookup will not work directly. Then probably one address is looked up and the others are not. 

  - I would look them up on the next cronjob.
  For not looked up address: 
    lookup -> wait (1+rand(0, 0+0.2) seconds) 
  Even if you upload 200 contacts the lookup will finish in less than 5min.

  - There is a window of time to next Cronjob + ~5min, where not all contacts are on the map, if the usere   uploads 200 contacts at once. 

I don't think that thinking about priorities will make sence for this window, even though I like the idea.

skjnldsv:
> * but Contacts does not support favorites yet, does it @nextcloud/contacts?

Nope :)

jancborchardt:
@tacruc and the retrieved lookup data is then stored in the Maps app, or where? Could make sense to store it in the contact instead, so other apps (like the Contacts app itself) can also use it.

tacruc:
It is stored for **now** at the maps app. It is a table of address and geo information. 
There are mainly three reasons, why I decided not to store it in the contact:

- Addresses which are in more than one contact are looked up multiple times,  the same person stored in different address books or form different users are different contacts and would be looked up multiple times.
- The code is executed on a vCard update event. Updating the vCard on it might have weird side effects.
  - One of them might is that you start typing the address in the contacts app, which sends it, while your still typing to the server, which looks it up and changes the vCard. You will than get an error while still typing that the vCard cloud not be saved, because it got changed externally.
  - This is actually connected to problem we have to solve: at the moment each address is looked up multiple times while typing and  I have no idea how to solve it.
- If the lookup would only be stored in the contact, where to store lookups without result.

This are the main reasons why it is for now in the maps app. Which by the way provides an Service to look up addresses, when we need it for events etc.

So as soon we agreed on a event, when we want to store the address in the contact, I'm happy to implement it. Some suggestions:

- When contact is displayed the first time at the map
- on some global user trigger, like safe position to contacts
- on a user trigger per contact
- on Cronjob
- ...?

eneiluj:
Hey! Nice brainstorm!

> each address is looked up multiple times while typing and I have no idea how to solve it

Maybe we should simply give up listening to vCard update events...OR find a way to put a delay before making the nominatim (or other) request and ask for coordinates for addresses which seems to be definitive in short term:

What about storing the information about which contact needs to be "placed" when the vcard is modified and let the cron job do the job :smile:. Frequency of this cron job would be "maximum" as we want it to be made soon and if there is nothing to do it does not cost anything.

There's still the 1s delay problem between nominatim requests but at least we don't try to make it several times for one contact. We could set a maximum number of requests to be made for each cron job to avoid having the cron job spending too much time on this.

I agree with @tacruc about storing coordinates in a Maps DB table. Anyway we can provide a simple way to save coordinates to the vcards afterwards from the interface.
We have to keep in mind that there's a big question to answer when there's an update of vcard address for a contact having a vcard address AND vcard geo: coordinates, we don't know if the user wants the new address to overwrite the vcard coordinates.

tacruc:
> Hey! Nice brainstorm!

Thx
> 
> > each address is looked up multiple times while typing and I have no idea how to solve it
> 
> Maybe we should simply give up listening to vCard update events...OR find a way to put a delay before making the nominatim (or other) request and ask for coordinates for addresses which seems to be definitive in short term:

The problem is not to big, as mentioned below there will be at most 1 request per second. 

> 
> What about storing the information about which contact needs to be "placed" when the vcard is modified and let the cron job do the job smile. Frequency of this cron job would be "maximum" as we want it to be made soon and if there is nothing to do it does not cost anything.
> 
> There's still the 1s delay problem between nominatim requests but at least we don't try to make it several times for one contact. We could set a maximum number of requests to be made for each cron job to avoid having the cron job spending too much time on this.
> 

The one second delay problem is already solved in the code. The lookup function will return null if there was an lookup in the last second. So in the cronjob we have to do something like wait 1second plus some random ammount. If we would wait one second in the function it would just delay all other code execution.

> I agree with @tacruc about storing coordinates in a Maps DB table. Anyway we can provide a simple way to save coordinates to the vcards afterwards from the interface.
> We have to keep in mind that there's a big question to answer when there's an update of vcard address for a contact having a vcard address AND vcard geo: coordinates, we don't know if the user wants the new address to overwrite the vcard coordinates.

I would add a new geo attribute.

As soon as the nextcloud dav system can handle multiple addresses. (I'm thinking of creating an issue or a PR). I would show the contact than multiple times on all provided locations (address and geo). Of course only once for a set of coordinates, which could come for an address and the corresponding geo tag.


So the open question remains, at which time do we want to save the geo information in

eneiluj:
Hey, I had a look at your code. I tried it and it works pretty well. Nice!

A few remarks:

* Could you add comments in `LookupMissingGeoJob.php` to explain what happens there? Is the job rescheduled when it fails or just if there were more than 200 contacts to localize?
* In `ContactsController.php`, what happens there? It needs more comments. What does the first IF means? Is this just getting coordinates for special contacts that are actually NC users?
* In `AddressService.php`, why are you using a file to store last lookup time? It could be stored in the database as app values like
```
$this->config->setAppValue('maps', $key, $value);
$value = $this->config->getAppValue('maps', $key);
* In `AddressService.php` again, in `getLastLookup()` I think it's safer to get the integer value from a string with `intval()` function
```
* The overall logic/principle/process of address lookup could be explained somewhere (like in `ContactsController.php`). It would be nice for future contributors to know what calls what and triggers what and when and how in a few sentences :wink:.

tacruc:
So a general remark on top:
the address service just cares for addresses it doesn't know contacts.

> Hey, I had a look at your code. I tried it and it works pretty well. Nice!
> 
> A few remarks:
> 
>    * Could you add comments in `LookupMissingGeoJob.php` to explain what happens there? Is the job rescheduled when it fails or just if there were more than 200 contacts to localize?

In both cases. So if it fails for a single address or if it looked-up 200 addresses but there where more not looked-up in the db.

> 
>    * In `ContactsController.php`, what happens there? It needs more comments. What does the first IF means? Is this just getting coordinates for special contacts that are actually NC users?

NC adds a vCard for the user itself. This vCard is skipped as it is in no addressbook.
 
>    * In `AddressService.php`, why are you using a file to store last lookup time? It could be stored in the database as app values like

I used the fileCache first, than I noticed that it didn't work. Is updating the config the right thing to do for a often and fast changing value? I think it is more appData https://docs.nextcloud.com/server/16/developer_manual/app/storage/appdata.html#appdata than an appConfig.


> * In `AddressService.php` again, in `getLastLookup()` I think it's safer to get the integer value from a string with `intval()` function
> 

Do you have any reference for this thought I couldn't find it.

> 
>    * The overall logic/principle/process of address lookup could be explained somewhere (like in `ContactsController.php`). It would be nice for future contributors to know what calls what and triggers what and when and how in a few sentences wink.

:+1:

eneiluj:
> Is updating the config the right thing to do for a often and fast changing value?

Accessing the DB is supposed to be much faster than accessing files. The database management system is supposed to give us abstraction to store/get data in an efficient way. It usually stores recent information in RAM and then optimize hard storage.

AppData uses the server filesystem and I think it's not very efficient.
It's your call. Anyway, to store/read such a small information, performance is not really an issue.

About `intval()`, I looked around and didn't find many arguments in favor. I guess I'm influenced by the memories of my strict algorithmic lessons. I would say if we're sure the casted value is an integer it's ok. The good thing about `intval()` is that it returns 0 if it fails to get an integer value. We potentially don't know what `(int)` does. Anyway, your call again.

Do you feel it's ready to be merged or do you want to add/change something?

tacruc:
I will adapt those changes and then we can merge

tacruc:
@eneiluj Now I would feel like merging.

PR #60, opened 2019-06-23
Rework eneiluj: bunch of new stuff

eneiluj:
Here are some new changes:

* rename device
* add start markers for tracks
* make 'occ maps:scan-photos' verbose
* add file actions to import compatible files as favorites or devices or tracks
* admin settings to setup routing engine (osrm-dem, osrm, graphhopper and mapbox)
* user option to choose routing engine and vehicle
* fix appstore makefile target

PR #61, opened 2019-06-26
Change class name to Version000009Date20190625000800

anaqreon:


anaqreon:
This is related to the discussion in https://github.com/nextcloud/maps/pull/60

eneiluj:
Well I'm almost sure if you install the app from scratch or if you just update the app this error won't show up.

I mean there is no mistake in naming it like it is in `rework-eneiluj`, your error is probably going to show up again if you do the exact same actions again.

Don't you think?

anaqreon:
Coincidentally I just followed the same procedure a few minutes ago on a completely separate host machine because I wanted to try the map on a real server to test the device tracking, and the app was unable to install (using the current branch). I can't look at the log file at the moment to ensure the error was identical but the use of Docker should make this behavior fairly consistent, right?

anaqreon:
I checked and it was the same error with the same solution to enable the app to install.

eneiluj:
Yep my bad, I didn't really read your changes. Thanks for the PR!

Issue #73, opened 2019-08-26
Favorites import/export from GPX is broken


anaqreon:
After commit c2b9c31f9d1b53b74cfcbcaa5557c6158cf4393b it appears that the import/export favorites using GPX files is broken. The errors I get are like the one pasted below, where it looks like there is a missing `endswith()` function, which was last modified in commit e88ce739fac30453371ad4faf6b30c9db8c992af, based on a search of the git log.

```
{
	"reqId": "Hk11oXsenZG0sR9BxlRs",
	"level": 3,
	"time": "2019-08-26T09:34:14+00:00",
	"remoteAddr": "1.2.3.4",
	"user": "andrew.manning",
	"app": "index",
	"method": "POST",
	"url": "/apps/maps/import/favorites",
	"message": {
		"Exception": "Error",
		"Message": "Call to undefined function OCA\\Maps\\Service\\endswith()",
		"Code": 0,
		"Trace": [
			{
				"file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
				"line": 166,
				"function": "importFavorites",
				"class": "OCA\\Maps\\Controller\\FavoritesController",
				"type": "->",
				"args": [
					"/Maps/2019-08-25 21_04_05 (-04_00) filtered-favorites.gpx"
				]
			},
			{
				"file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
				"line": 99,
				"function": "executeController",
				"class": "OC\\AppFramework\\Http\\Dispatcher",
				"type": "->",
				"args": [
					{
						"__class__": "OCA\\Maps\\Controller\\FavoritesController"
					},
					"importFavorites"
				]
			},
			{
				"file": "/var/www/html/lib/private/AppFramework/App.php",
				"line": 126,
				"function": "dispatch",
				"class": "OC\\AppFramework\\Http\\Dispatcher",
				"type": "->",
				"args": [
					{
						"__class__": "OCA\\Maps\\Controller\\FavoritesController"
					},
					"importFavorites"
				]
			},
			{
				"file": "/var/www/html/lib/private/AppFramework/Routing/RouteActionHandler.php",
				"line": 47,
				"function": "main",
				"class": "OC\\AppFramework\\App",
				"type": "::",
				"args": [
					"FavoritesController",
					"importFavorites",
					{
						"__class__": "OC\\AppFramework\\DependencyInjection\\DIContainer"
					},
					{
						"_route": "maps.favorites.importFavorites"
[...]

PR #75, opened 2019-08-26
Fix "endswith" declaration and remove contact groups colors

eneiluj:
There was a huge problem with a little function declared multiple times. It was not detected wen running the tests or loading the main app page...

I also removed the colors for contact groups as it's probably useless. I've put the main theming color for all group icons instead. We could also set their color to black/grey.

As you wish.


jancborchardt:
> I also removed the colors for contact groups as it's probably useless. I've put the main theming color for all group icons instead. We could also set their color to black/grey.

I’d indeed say let’s use simple black so it’s uniform. :) Otherwise good!

eneiluj:
A'ight. I agree. It's now black.

jancborchardt:
@anaqreon also check this out since it should fix the issue you reported at https://github.com/nextcloud/maps/issues/73 :)

eneiluj:
@anaqreon I'll check that tomorrow. Thanks for the quality of your tests :smile:.

eneiluj:
@anaqreon I did a quick fix to favorite export procedure. Could you export and try to import again?

Thanks again for finding this bug.

eneiluj:
Btw, it's still far from perfect. Characters like "<" and ">" might be problematic. Export is very naive and should be converted to something using a proper library to write correct and clean XML.

eneiluj:
Yep it really needs a clean implementation. I'll convert "<" and ">" for the moment but it will be much safer by writing with a library.

PR #76, opened 2019-08-27
Update readme, info.xml, add screenshots

jancborchardt:
🚨 Getting ready for release!

- Some nice screenshots which illustrate the general view, photos on the map, and routing. These are also used for the app store. (I left the 2 existing screenshots in for nostalgia :)
- Updated the readme with short feature description and development setup.
- Also updated the info.xml with info according to the readme.

Please review @nextcloud/maps


jancborchardt:
@anaqreon the readme is modeled after what we use for other apps like Text, Social, etc., so kind of using the standard. :)

The screenshots are also not about showing everything (especially not at once), but giving a general feel. Ppl should just get "huh this is is like Googe Maps plus my own data". :)

I do agree it would be nice to have them on a smaller screen, but there's a  problem:
- When we take them with a smaller screen / browser window, they are blurry on bigger
- My usual strategy of zooming in 150% doesn't work cause the map tiles and text are very blurry then

That's why I just went for this.

jancborchardt:
@anaqreon can you open a separate issue for the import issue? cc @tacruc @eneiluj

appinfo/info.xml
```diff
@@ -3,18 +3,37 @@
...
+    <website>https://github.com/nextcloud/maps</website>
+	<bugs>https://github.com/nextcloud/maps/issues</bugs>
+	<repository type="git">https://github.com/nextcloud/maps.git</repository>
+    <screenshot>https://raw.githubusercontent.com/nextcloud/maps/master/screenshots/screenshot1.png</screenshot>
+	<screenshot>https://raw.githubusercontent.com/nextcloud/maps/master/screenshots/screenshot2.png</screenshot>
+	<screenshot>https://raw.githubusercontent.com/nextcloud/maps/master/screenshots/screenshot3.png</screenshot>
     <dependencies>
         <lib>exif</lib>
-        <nextcloud min-version="15" max-version="17"/>
+        <nextcloud min-version="15" max-version="18"/>
```
nickvergessen: Nope, dont pretend to be compatible with a non-existing version

jancborchardt: Just doing this cause as soon as 17 is out, which is very soon, the local dev instances will be 18.

jospoortvliet:
It looks very nice!
One minor comment - I'd not mention the super nerdy GNOME Maps, though, instead say Desktop apps perhaps - nobody knows GNOME Maps. Marble at least is available for Windows and Mac OS X so it likely has way more users but even that is a tiny user base, I bet. Just say Desktop - and then we can look what desktop mapping apps people actually use, if any.

appinfo/info.xml
nickvergessen: But versions with 18 max should not be published to the appstore yet

jancborchardt:
@jospoortvliet I mentioned GNOME Maps specifically cause they are the only ones at this point where there’s specific contact – see the linked issue. :) And I think it’s good to keep them, so the others see that it’s valuable. (Also I ♥ GNOME)

appinfo/info.xml
eneiluj: Thanks @nickvergessen . Done.

PR #78, opened 2019-08-28
fix file hooks

eneiluj:
I had removed `<types>` in info.xml because the `occ app:check-code` complained about it and I didn't find anything in the doc. It appears that file hooks need that to work.

It's now working again.

I also moved FileHooks class to another namespace to make it like other apps and like advised in the doc.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -48,4 +48,7 @@
             <route>maps.page.index</route>
         </navigation>
     </navigations>
+    <types>
+        <filesystem/>
+    </types>
 </info>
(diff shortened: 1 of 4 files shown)


jancborchardt:
How could I test this? :)

eneiluj:
Just check if you see a track file listed in Maps app after having added it to your files.

appinfo/info.xml
```diff
@@ -48,4 +48,7 @@
             <route>maps.page.index</route>
         </navigation>
     </navigations>
+    <types>
+        <filesystem/>
+    </types>
```
tacruc: But now you include types again? I thought you want it to be removed?

tacruc: Ah now I get it, you removed it on "accident"

PR #79, opened 2019-08-28
run photo and track scan on app installation

eneiluj:
We can run the scan with an install repair step. I've tested it on a fresh Nextcloud 16 install. I added a few photos and track files, then installed Maps app and it works :smile:.

Downside of this method, i guess an admin can think there is some kind of failure if the installation is too long.

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -179,6 +179,14 @@
                 async: true
             }).done(function (response) {
                 optionsValues = response.values;
+
+                // check if install scan was done
+                if (!optionsValues.hasOwnProperty('installScanDone') || optionsValues.installScanDone !== 'yes') {
+                    OC.Notification.showTemporary(
+                        t('maps', 'Media scan was not done yet. Wait a few minutes/hours and reload this page to see your photos/tracks.')
+                    );
+                }
+
                 // set tilelayer before showing photo layer because it needs a max zoom value
                 if (optionsValues.hasOwnProperty('displaySlider') && optionsValues.displaySlider === 'true') {
                     $('#timeRangeSlider').show();
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -35,6 +35,11 @@
         <lib>exif</lib>
         <nextcloud min-version="15" max-version="17"/>
     </dependencies>
+    <repair-steps>
+        <install>
+            <step>OCA\Maps\Migration\InstallScan</step>
+        </install>
+    </repair-steps>
     <commands>
         <command>OCA\Maps\Command\RescanPhotos</command>
         <command>OCA\Maps\Command\RescanTracks</command>
--- a/lib/BackgroundJob/LaunchUsersInstallScanJob.php
+++ b/lib/BackgroundJob/LaunchUsersInstallScanJob.php
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * Nextcloud - maps
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Julien Veyssier
+ * @copyright Julien Veyssier 2019
+ */
+
+namespace OCA\Maps\BackgroundJob;
+
+use \OCP\BackgroundJob\QueuedJob;
+use \OCP\BackgroundJob\IJobList;
+use \OCP\AppFramework\Utility\ITimeFactory;
+use OCP\IUserManager;
+use OCP\IUser;
+
+use \OCA\Maps\BackgroundJob\UserInstallScanJob;
+
+class LaunchUsersInstallScanJob extends QueuedJob {
+
+    private $jobList;
+
+    /**
+     * LaunchUsersInstallScanJob constructor.
+     *
+     * A QueuedJob to launch a scan job for each user
+     *
+     * @param IJobList $jobList
+     */
+    public function __construct(ITimeFactory $timeFactory, IJobList $jobList, IUserManager $userManager) {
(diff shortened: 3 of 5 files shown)


jancborchardt:
cc @rullzer @MorrisJobke @nickvergessen can you check if this a good way to do it? :) The goal is to check through people’s files for photos and tracks to be displayed on the map directly on first run.

rullzer:
Yeah that won't work. For example if I install the app on my personal instance now it will have to go over 500 gb of pictures.

Also a big chance this will time out on several systems. 

What you could do is insert a background job in the repair step.

This background job then does what you want. And will be run by cron.

Then I would also propose to have some config var that gets set when the initial scan is done. So you can show a message "your initial scan is not complete yet"

tacruc:
@rullzer is timeout a problem for background jobs? So the initial scan be in multiple background jobs or is one big job running multiple hours fine in an background job?

rullzer:
> @rullzer is timeout a problem for background jobs? So the initial scan be in multiple background jobs or is one big job running multiple hours fine in an background job?

It depends. If people run it from cron then it is fine.
If people run it from ajax there is still the timeout problem.

What we usually do in server is a layered system.

1. Repair step inserts backgroundjob 1
2. backgroundjob1 loops over all seen users and for each inserts bacgroundjob2
3. bakgroundjob2 does the actual scanning for that user.

That way we make sure that one huge job doesn't stop the queue from progressing.

eneiluj:
@rullzer Thanks a lot for the advice. I'm on it!

eneiluj:
@rullzer Why not inserting all bakgroundjob2 directly from the install repair step?

rullzer:
> @rullzer Why not inserting all bakgroundjob2 directly from the install repair step?

Consider a system with >20k users

eneiluj:
Yep a'ight :smile:.

eneiluj:
Ok done. The repair step schedules a "job1" which schedules a "job2" for each user. A "job2" scans photos and tracks for a user.

I've tested it with a small amount of users and files. It works with Ajax cron task.

User is warned if the media scan was not done for him/her yet.

jancborchardt:
I disabled the app and enabled it again to test (that works, right?). It shows the notification still:

> Media scan was not done yet. Wait a few minutes/hours and reload this page to see your photos/tracks.

I only have a few photos (15 already shown and I think those are all geotagged ones) and 0 tracks atm. So it should not show the note anymore?

eneiluj:
@jancborchardt It depends the way the cron task is triggered on your NC instance. If it's with Ajax, there might be lots of other jobs to launch before the ones scheduled by Maps. If it's with a system cron job, you have to wait for it to be triggered twice because first there is a job scheduling one job per user and then those jobs are executed.

eneiluj:
You can check if something is scheduled with:

```
sqlite3 /var/www/html/test/nextcloud/data/owncloud.db 'select * from oc_jobs;' | grep -i maps
```

assuming you use SQLite and you have access to the DB file.

js/script.js
```diff
@@ -179,6 +179,14 @@
                 async: true
             }).done(function (response) {
                 optionsValues = response.values;
+
+                // check if install scan was done
+                if (!optionsValues.hasOwnProperty('installScanDone') || optionsValues.installScanDone !== 'yes') {
+                    OC.Notification.showTemporary(
+                        t('maps', 'Media scan was not done yet. Wait a few minutes/hours and reload this page to see your photos/tracks.')
```
tacruc: minutes/hours I think just mentions hours is fine here.

jancborchardt:
Ok – @rullzer could you check if the code/mechanism is proper? :)

jancborchardt:
@eneiluj and from the database I get this:
```
321|OCA\Maps\BackgroundJob\LookupMissingGeoJob|[]|0|1566941482|0|0
```
Do the 0s mean it’s done? :D

eneiluj:
It's not the install scan job there. This one is independent and dedicated to contact addresses.

Either the install job was not scheduled when you disabled/enabled the app or it has been executed.

What about

```
sqlite3 /var/www/html/test/nextcloud/data/owncloud.db 'select * from oc_maps_photos;' | wc -l
```

eneiluj:
It's much simpler to test this with a fresh NC install to make sure there is nothing in Maps DB before testing this branch.

jancborchardt:
@eneiluj the other command just outputs "35".

Yeah, as said I might not be the best person to test this as this needs a proper code review. So @rullzer or maybe @ChristophWurst @nickvergessen could briefly check. :)

eneiluj:
@jancborchardt So there are 35 photos found by Maps app. If you started with an empty Nextcloud instance, this means the scan was done after Maps installation. If you tested with existing data in Maps DB, this does not show if install scan works :wink:.

eneiluj:
Thanks @tacruc . I'll still wait a few minutes/hours (:wink:) for a confirmation that I did it properly technically speaking. Then I'll merge and produce the v0.1.0 archive!

nickvergessen:
Looks good in general, but I guess the scanning will fail for bigger users with ajax cron, but well that's their tough luck. Maybe a console command to rescan a user can help in that case.

eneiluj:
Thanks @nickvergessen. There already are two commands, one to scan photos and the other to scan tracks.

js/script.js
jancborchardt: I’d say let’s keep minutes/hours cause "hours" sounds a bit harsh, no? :D

PR #80, opened 2019-08-28
Mention bias of Mercator map projection, thanks @alicetragedy

jancborchardt:
Making a maps app this is an important thing to note, so it’s added both in the README and in the settings of the app itself. Thanks @alicetragedy for reminding!

mercator bias

Please review @nextcloud/maps 

(And yes, it is ok to have an external link in the app, we do it similarly in Mail where we talk about how to encrypt your mail with Mailvelope. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@
 - **📱 Devices:** Lost your phone? Check the map!
 - **〰 Tracks:** Load GPS tracks or past trips. Recording with [PhoneTrack](https://f-droid.org/en/packages/net.eneiluj.nextcloud.phonetrack/) or [OwnTracks](https://owntracks.org) is planned.
 
+Future plans:
+- **📆 Events on the map:** Know where you need to go next!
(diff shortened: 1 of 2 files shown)


anaqreon:
This subject obviously has social and political implications, so it might be better to make the statement more general and not include specific links. For example, someone in the southern hemisphere might feel offended that the "north pole is upward" convention is used. Who knows what else might be offensive to others. Perhaps a general statement about how the particular projection used is not driven by ideology but more simply by available free and/or open source data, and that more options will be supported as they are available.

jancborchardt:
I think we should be specific, especially because it has social and political implications. "The True Size of Africa" is a good specific example to make people aware of the systemic issue instead of just doing some generalized statement about how we are just using what’s out there – which I think is a cop-out.

(Also it kind of implies the preference of the northern hemisphere on a specific example. And well, the whole continent of Africa is very specifically most marginalized, so it’s proper to bring it up.)

PR #81, opened 2019-08-29
avoid showing sidebar when grabing the map to the right

eneiluj:
Solution given by @skjnldsv works like a charm.


eneiluj:
@anaqreon @skjnldsv Thanks!

jancborchardt:
Goood stuff @eneiluj and thanks a lot @skjnldsv! :)

Issue #56, opened 2019-05-05
TODO before first release


eneiluj:
Here is an extensible list of what's missing before we can think about making a first release:

## General
* [x] get a certificate to publish to appstore
* [x] fix/test makefile "appstore" target
* [x] add filtypes actions to import favorites/devices from gpx/kml/kmz
* fix CI script:
    * [ ] at some point test complained it couldn't create user folder
    * [ ] autoloader problem (are we doing something wrong when installing NC?)
* [x] fix second migration script which does not declare auto increment field as primary key
* [x] make some tests (at least PhpUnit)
* [ ] use Webpack to manage JS dependencies
* [x] make time slider optional
* [ ] use a library to write valid XML when exporting things

## Routing
* [x] implement admin options to setup routing: choose routing engine (Osrm/Graphhopper...), service URL, API key, routing type (car, bike, foot...)
* [x] let the user choose routing type in the UI (if declared as possible in admin settings)
* [ ] add routing button to export track (was in sidebar before)

## Contacts
* [x] display contacts that have an address #34 

## Favorites
* [ ] have at least one client able to sync with us:
    * [ ] Marble would require a few changes to be compatible
    * [ ] Gnome Maps might become compatible this summer
    * [ ] OsmAnd would be nice! Nothing like that yet.

## Tracks
* [x] add a marker with an icon where the tracks begin
* [x] list all tracks, use file hooks to populate track list just like it's done for photos
* [x] make the track list sortable by filedate/filename
* [x] make tracks searchable by file name

## Devices
* [x] clarify device identification (currently: user agent is the ID)
* [x] if we keep user agent as ID, parse it better (avoid browser version number for example)
* [x] make PhoneTrack-Android able to log to Maps

## Photos
* [ ] improve what happens when clicking on a photo (use gallery viewer?)
* [ ] handle group sharing events/hooks
* [x] make `occ maps:scan-photos` verbose
* [x] scan photos after app installation
* [x] hide non-localized photos feature for now

e-alfred:
Awesome roadmap! The only thing missing is probably the verbose output of `occ maps:scan-photos` which doesn't show anything if run but actually works if you check the database.

eneiluj:
@e-alfred Yep thanks for the remark. It's been added to the list.

jancborchardt:
> have at least one client able to sync with us

I would say that this likely comes after the first release, not right out of the gate. :)

jancborchardt:
[...]

PR #82, opened 2019-08-29
Fix search box position on mobile with navigation menu, ref #56

jancborchardt:
Fixed the issue reported by @tacruc:
Desktop, still working as now:
desktop search

Mobile is fixed:
mobile search


eneiluj:
Nice!

Could you make it work when routing div is displayed?

search

jancborchardt:
Yup, fixed :)

Issue #85, opened 2019-08-29
Undefined index: FN 


tacruc:
Second error:
`
 Undefined index: FN at /srv/www/nextcloud/apps/maps/lib/Controlle/ContactsController.php#101
`

eneiluj:
Wow, I didn't know it was possible FN is not defined. Do you have a contact without name?

tacruc:
It might well be, as I saved sum numbers in android without specifying a name, but I have to drive home and check the db for a exact answer.

eneiluj:
Ok anyway the fix is quite easy.

PR #87, opened 2019-08-29
refs #85 fix bug with unnamed contacts

eneiluj:
I guess this should do it.

### DIFF
--- a/lib/Controller/ContactsController.php
+++ b/lib/Controller/ContactsController.php
@@ -70,7 +70,7 @@ class ContactsController extends Controller {
                     $geo = $c['GEO'];
                     if (strlen($geo) > 1) {
                         array_push($result, [
-                            'FN'=>$c['FN'],
+                            'FN'=>$c['FN'] ?? '???',
                             'URI'=>$c['URI'],
                             'UID'=>$c['UID'],
                             'ADR'=>'',
@@ -98,7 +98,7 @@ class ContactsController extends Controller {
                         }
                         if (strlen($geo) > 1) {
                             array_push($result, [
-                                'FN'=>$c['FN'],
+                                'FN'=>$c['FN'] ?? '???',
                                 'URI'=>$c['URI'],
                                 'UID'=>$c['UID'],
                                 'ADR'=>$adr->getValue(),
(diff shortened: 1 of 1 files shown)


tacruc:
Maybe not ??? but an empty string?

eneiluj:
Well I don't want to face other problems like generating an avatar image for an empty string etc...

tacruc:
Ok

tacruc:
I found why there is no FN attribute, the cards just have the N attribute.

eneiluj:
Wait, shouldn't we read 'N' when 'FN' is not set? What does the 'N' field contain?

eneiluj:
What do you think about something like
```
'FN'=>$c['FN'] ?? $c['N'] ?? '???',
```
?

tacruc:
```
private function N2FN(string N) {
  return explode(N,';')[3]+explode(N,';')[1]+explode(N,';')[0]
}

'FN'=>$c['FN'] ?? $this->N2FN($c['N']) ?? '???',
```

eneiluj:
I can't understand what you wrote unless you explain what's inside 'N' field.

tacruc:
`Doe;John;;Dr;`

eneiluj:
Ok and do you know when/why 'N' is set and 'FN' is not?

tacruc:
Its an address book I created by importing an CSV file with kaddressbook. It might be an error in the import process.

jancborchardt:
Maybe @skjnldsv from @nextcloud/contacts can help with that Contacts-related question? :)

eneiluj:
Well then `N2FN` could be more secure by verifying the explode gives at least 4 elements. It could be called once and put in a variable too.

Do we agree on something like:

```
private function N2FN(string $n) {
  if ($n) {
    $spl = explode($n,';');
    if (count($spl) >= 4) {
      return $spl[3] + $spl[1] + $spl[0];
    }
    else {
      return null;
    }
  }
  else {
    return null;
  }
}

'FN'=>$c['FN'] ?? $this->N2FN($c['N']) ?? '???',
```
?

tacruc:
Looks good, we probably need to add some spaces between the parts.

eneiluj:
Could you give it a little try?

jancborchardt:
Briefly tested and seems good, but I can’t judge the code.

PR #91, opened 2019-08-29
Catch PelExeptions, for example on broken images

tacruc:
The photo scan crashes on broken images, because of not catch PelExeptions.


eneiluj:
Just to be sure: Do you think this PR solves #96 and #97?

tacruc:
Yes

PR #103, opened 2019-08-30
Added different mapbox profiles

tacruc:
Added routing with bike and by foot for mapbox.

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -835,7 +835,11 @@
                 delete this.routers.osrmDEMO;
             }
             if (optionsValues.hasOwnProperty('mapboxAPIKEY') && optionsValues.mapboxAPIKEY !== '') {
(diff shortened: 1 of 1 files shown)


eneiluj:
Did you try this in real conditions with an actual Mapbox API key?

tacruc:
Yes, is it not working for you?

eneiluj:
I don't have a Mapbox API key :smile:.

I was just wondering if you added this after having red the doc and if you had the ability to try it for real.

tacruc:
Well I just registered and got an api key. Is a thing of 5 minutes. Than I read the docs.

eneiluj:
After deploying an OSRM server and a GraphHopper one, I became lazy enough not to try Mapbox even if it's quick :smile:.

tacruc:
I didn't get osrm working, as I would need the mapping data for Europe, where it says that it needs arround 150GB of ram. Thats why I went for Graphhopper and mapbox.

Issue #135, opened 2019-09-05
The right-click routing-options don't work


szaimen:
After setting up routing over the admin-settings, when right-clicking on the map, additional routing-options appear. But they don't work for me. (I click on the buttons, but nothing happens.
I am using a chromium-based browser on windows, if that is important.
![image](https://user-images.githubusercontent.com/42591237/64382270-8a153300-d035-11e9-9952-023de236f678.png)

WNYmathGuy:
I just mentioned this too in https://github.com/nextcloud/maps/issues/110#issuecomment-528605274

eneiluj:
I just pushed a fix in `fix-issue135` branch if you feel like testing it. Thanks for the bug report!

WNYmathGuy:
P.S. @eneiluj I know this is closed, but...
I'm currently running the Maps 0.1.2 and just wondered if this fix in 0.1.3 also dealt with the **Your devices** proximity circle shielding the map from any possible right-click or left-click attempts? If not I'll be glad to raise a new issue for you.

PR #150, opened 2019-09-15
refs #135 fix context menu routing actions

eneiluj:
It was a big mistake. Context is different when calling function from map elements. `this` was referring to the map object and the code was written as if `this` was the routing controller.

I think it's fixed now.


tacruc:
Tested

PR #152, opened 2019-09-15
Fix #128 and generally improve initial media scan

### LINKED ISSUE #128: "Media scan was not done yet" .. ?
I just installed the app and I'm not familiar with it that much yet. But when I run & access it, this is what I see:
```
Media scan was not done yet. Wait a few minutes/hours and reload this page to see your photos/tracks.
Failed to load contacts

```
It's been 10+ hours and nothing has changed. I also tried running OCC scan.

The log outputs additional error:

Screenshot_1

I use Nextcloud (15) snap package.

eneiluj:
This fixes multiple aspects of #128.

* minimum NC version is now 16 because there is a call to `OC\ContactsManager::getUserAddressBooks()` which appeared in 16 as described [here](https://docs.nextcloud.com/server/16/developer_manual/api/OC/ContactsManager.html?highlight=contactsmanager#OC%5CContactsManager::getUserAddressBooks)
* user `installScanDone` value is set to "yes" after occ track or photo scan
* user `installScanDone` value is initialized to "no" on Maps app install
* interface media scan warning is displayed only if `installScanDone` value is set and equals "no"

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -183,7 +183,7 @@
                 optionsValues = response.values;
 
                 // check if install scan was done
-                if (!optionsValues.hasOwnProperty('installScanDone') || optionsValues.installScanDone !== 'yes') {
+                if (optionsValues.hasOwnProperty('installScanDone') || optionsValues.installScanDone === 'no') {
                     OC.Notification.showTemporary(
                         t('maps', 'Media scan was not done yet. Wait a few minutes/hours and reload this page to see your photos/tracks.')
                     );
--- a/js/script.js
+++ b/js/script.js
@@ -183,7 +183,7 @@
                 optionsValues = response.values;
 
                 // check if install scan was done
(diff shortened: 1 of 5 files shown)


js/script.js
```diff
@@ -183,7 +183,7 @@
                 optionsValues = response.values;
 
                 // check if install scan was done
-                if (!optionsValues.hasOwnProperty('installScanDone') || optionsValues.installScanDone !== 'yes') {
+                if (optionsValues.hasOwnProperty('installScanDone') || optionsValues.installScanDone === 'no') {
```
tacruc: I don't get this change, if optionValues has the property, it is true and the value will not be checked at all. If the optionsValues does not have the property it is going to check the Value, which will lead to an error?

tacruc: should there be an &&?

eneiluj: Yes indeed, doing too many things at the same time. Thanks for the quality review!

eneiluj:
Ok now the condition makes sense :wink:.

Issue #132, opened 2019-09-05
Popups are out of place on contacts with default picture


szaimen:
When hovering with the mouse over one contact with default pictures, the popup that appears is out of place.
See the screenshot:
![image](https://user-images.githubusercontent.com/42591237/64356608-6d600780-d003-11e9-9f1e-d2caa15a287d.png)
(The picture-bubble is hidden behind the popup.)
I am using a chromium based Browser on Windows if that is important.
Thank you!

jancborchardt:
@Cancan68 this seems like a different problem. Could you open a separate issue for that? :) Thanks!

tacruc:
![image](https://user-images.githubusercontent.com/402891/64924169-d56fd400-d7e1-11e9-87da-0d9a7db47957.png)
Is this how it is supposed to be?

eneiluj:
Who are you asking to?

My answer is: yes :wink: The tooltip is now placed at the center of the marker.

tacruc:
I would have assumed that the triangles should point to the same place.

eneiluj:
Well if we do like that marker becomes invisible when tooltip is visible. I don't like that but let's talk about it.

tacruc:
@eneiluj I see that why I ask if this placement was on purpose. 
@jancborchardt  what do you think about the placement?

jancborchardt:
I’d agree with @tacruc, the triangles should point to the same place. Basically it should look as if the marker just "grew". Makes sense? :)

anaqreon:
@jancborchardt I attempted to address your concern with a simple commit to pull request #154. I made some mistakes with the git branches so hopefully I didn't mess things up too badly.

jancborchardt:
The original issue was fixed through https://github.com/nextcloud/maps/pull/154, the design followup is in https://github.com/nextcloud/maps/pull/195 :)

PR #154, opened 2019-09-15
refs #132 fix popup shift when avatar appears late

eneiluj:
It should solve the problem mentioned in #132 for all browsers. If the image size is set before it has been loaded, popup size does not change after having been displayed and the shift does not happen.

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -593,6 +593,8 @@ tr.selected td {
 .tooltip-contact-avatar {
     min-width: 70px;
     min-height: 70px;
+    max-width: 70px;
+    max-height: 70px;
     margin-left: 40px;
     margin-right: 40px;
     border-radius: 50%;
--- a/css/style.scss
+++ b/css/style.scss
@@ -591,7 +591,8 @@ tr.selected td {
 }
 
 .tooltip-contact-avatar {
-    width: 70px;
+    min-width: 70px;
+    min-height: 70px;
     margin-left: 40px;
     margin-right: 40px;
     border-radius: 50%;


anaqreon:
I can confirm that this pull request does fix the issue in the Chromium web browser (which is not a problem in Firefox). The popup position is centered on the contact marker image; it is not coincident with the contact marker as of commit 4439fd5d3d07a7dfc34f0570134fb95b83c0620c.

jancborchardt:
Added a commit which fixes the popup being too big for contacts with a profile image (cause it just sized it up to whatever size the contact image had, since the width limit was missing). :) Now it works in both Chrome and Firefox.

Anything about the js/contactsController.js conflict? Will see if I can resole it.

jancborchardt:
@anaqreon I did the conflict resolution, but now for some reason your change with moving the popover doesn’t work anymore, even though as far as I can see I took it over 1 to 1. Sorry – do you have any idea what’s wrong there?

jancborchardt:
@eneiluj @tacruc the line that @anaqreon changed was removed completely in master, so I assume something changed regarding how to position the popup?

eneiluj:
As far as I can see, @anaqreon's commit doesn't have any effect because contact tooltip shows/hides thanks to `onContactMouseover` and `onContactMouseout` methods. There should be no tooltip bind. It's done that way to avoid showing the tooltip when the popup is there.

I guess @anaqreon didn't pull fix-issue132 before working or I don't quite understand what happened.

Anyway, I feel we're mixing two topics here.

* wrong popup/tootlip placement when avatar appears too late or is too big (fixed by @jancborchardt and @eneiluj). It was because the tooltip/popup size was changing after it was displayed, making it move.
* choose whether contact tooltips/popups should be placed at the exact marker place or at its center. Personally I prefer showing it at the center to better express the fact it's related to the contact marker which stays visible. Otherwise tooltip/popup hide the marker and it feel weird.

Bottom line: I think we could keep @jancborchardt's commit and get rid of @anaqreon's, then merge this PR, then ask ourselves where to place tooltips/popups. What do y'all think?

eneiluj:
To make it clearer, I removed @anaqreon's commit and pushed again `fix-issue132`. It looks good to merge to me.

anaqreon:
Wow, I have no idea how I screwed up that simple change so royally. Sorry guys.

eneiluj:
@anaqreon Don't worry, I'm pretty sure everybody messed up with git at least a thousand times :wink:.

jancborchardt:
> Bottom line: I think we could keep @jancborchardt's commit and get rid of @anaqreon's, then merge this PR, then ask ourselves where to place tooltips/popups. What do y'all think?

Sounds good! :)

And as said in the other issue, I do think @anaqreon’s change was good, as it just looks like the marker is growing. (Whereas currently it’s a marker on a marker, which is weird. :D)

jancborchardt:
@anaqreon so if you want to open another pull request for the position of the popover after click, that’d be cool! :)

PR #167, opened 2019-09-18
Use the backend to perform routing API requests

aszlig:
With this, the routing API requests are no longer performed via the user's browser but are sent to the backend where the API calls are proxied accordingly.

I've done this for two reasons:

  * First of all, doing the API requests by the server makes it harder for external services to track user behaviour, because all requests are coming from a single source.
  * Second, the API keys are now no longer exposed to users.

I've tested this with OSRM, Mapbox and GraphHopper.


lib/Http/ProxyResponse.php
```diff
@@ -0,0 +1,76 @@
...
+    public function __construct(string $url) {
+        $this->url = $url;
+    }
+
+    /**
+     * Send the API request to the given URL and set headers for our response
+     * appropriately.
+     */
+    public function sendRequest(ILogger $logger): bool {
+        if (($curl = curl_init()) === false) {
```
aszlig: Hm, actually found that there is `\OCP\Http\Client`, which uses Guzzle. I think that's more preferable to this, as it's used by other functionality in Nextcloud. Will replace this during the weekend.

jancborchardt:
@nextcloud/maps anyone with more technical knowledge than me is welcome to do a review of this. :)

@aszlig thanks a lot for your contribution and sorry for the wait!

aszlig:
@jancborchardt: No need to be sorry for anything, got PRs that took years to integrate :-) Besides, I'm also a newbie if it comes to the Nextcloud API.

jancborchardt:
@thomas-mc-work @Galbar since @tacruc @eneiluj are a bit busy atm, and you opened pull requests for Maps too (thanks! :) it would be awesome if you could review this too. Thanks! :tada:

lib/Controller/UtilsController.php
```diff
@@ -79,21 +79,30 @@ public function getOptionsValues() {
...
+                if ($this->config->getAppValue('maps', $opt) === '') {
+                    continue;
+                }
+                $features[$feature] = true;
+                break;
+            }
         }
-        return new DataResponse(['values'=>$ov]);
+
+        return new DataResponse(['values' => $ov, 'features' => $features]);
```
eneiluj: Did you call it "features" because you think ahead and predict it can be used for other types of features? I wonder if it could be named more explicitly like "routingProviders" and leave other types of features in another key to make it clearer. I mean "routing" is a feature but "mapbox" is not...

If you agree with me, let's make it more explicit anywhere "feature" is used.

aszlig: Yes, this was meant to be used in a generic way, eg. for also serving tiles via a local backend, but yeah, probably makes sense to name this `routingProviders` and if we have tile support we can introduce a new `tileProviders` and whatnot.

aszlig:
> Maybe those two remarks don't make much sense because routing requests are very small and the performance loss will almost never be felt.

I haven't made any large-scale benchmarks but I did deploy this on a range of Nextcloud instances already, which only have a small amount of users (~50-70). For larger installations, we can still add caching if this really turns out to be a performance bottleneck, because as you mentioned, those requests are usually very infrequent and the requests themselves are pretty small (the replies could be a bit longer, but only if you try to walk to Australia from Germany for example).

> Well, to conclude wink, let me try this out and play with it, let's wait for @tacruc feedback and from someone with deeper knowledge of how such things are generally done in Nextcloud (@jancborchardt any idea?).

If you look at an app like `calendar`, you'll find that it's done in a similar way:

https://github.com/nextcloud/calendar/blob/924bad1ec44cf20b0ddc95502901afeac3fa5ab6/controller/proxycontroller.php
> 
> I'll be glad to approve this in (much) less than a year grin.

:tada:

aszlig:
Another app (`weather`) also does it in [a similar way](https://github.com/nextcloud/weather/blob/566a955ce57dc5fc19a17070b56b919b528b94a0/lib/Controller/IntermediateController.php) for their API request, but uses `curl` directly as I did [prior to realizing that there already is a Guzzle wrapper in Nextcloud](https://github.com/nextcloud/maps/pull/167/commits/843cfc417e1d633fdd55f5ac213d5b6552486b06).

jancborchardt:
> let's wait for @tacruc feedback and from someone with deeper knowledge of how such things are generally done in Nextcloud (@jancborchardt any idea?).

I would say @rullzer, and since @aszlig mentioned the Calendar app too, @georgehrke. :)

georgehrke:
Please don't take example in the ProxyController in the calendar app. 🙈
It's a piece of really crappy code that will be removed very soon.

---

Just some quick remarks about this PR (not necessarily complete)
- Please use rate-limiting (Read up about the @UserRateThrottle annotation, e.g. @UserRateThrottle(limit=5, period=100))
- Why does it have a `@NoCSRFRequired` annotation?

aszlig:
> Please don't take example in the ProxyController in the calendar app. :see_no_evil:
> It's a piece of really crappy code that will be removed very soon.

I didn't use it as an example for writing the code here, but was looking for other examples doing it in a similar way after @eneiluj's comment.

> 
> Just some quick remarks about this PR (not necessarily complete)
> 
>     * Please use rate-limiting (Read up about the @UserRateThrottle annotation, e.g. @UserRateThrottle(limit=5, period=100))

Good point, will implement this.

>     * Why does it have a `@NoCSRFRequired` annotation?

Because I didn't yet dig into Leaflet enough to pass the token along. Right now the attack scenario for this is that an attacker could exhaust API limits for routing services, eg. by directing logged-in user's browser to mass-visit something like `https://nextcloud.example.org/apps/maps/api/requestRoute/graphhopper?...`. Will dig into that this weekend.

eneiluj:
Hey, there will be a conflict between this PR and #187 because I get the Mapbox API key from option values to create Mapbox tile layers in the UI.

I don't see how we can easily use the backend to perform tile request... And most importantly I don't know if it would be a good idea to have all the tile-related traffic through the server, performance-wise.

So I guess the conflict is quite big: #187 "forces" us to expose the Mapbox API key in the UI code. Sorry about that (but vector tiles are awesome):smile:.

@aszlig What do you think?

georgehrke:
> Because I didn't yet dig into Leaflet enough to pass the token along.

Ideally, all the usage of jQuery `$.ajax({` would be replaced with our [`@nextcloud/axios` package](https://www.npmjs.com/package/@nextcloud/axios).

If you use that package, it will automatically include the request token necessary to pass the CSRF check in every request.

aszlig:
@eneiluj: I think proxying (and maybe caching?) Mapbox tiles is something we could do as part of another PR, because after all this PR is about routing. Unfortunately, I did not manage to work on this during the weekend, but I'll try resume (and rebase) on this later this week.

aszlig:
Since I didn't get to work on that since quite a while, I at least rebased this branch against current master.

What's still missing is rate limiting and the removal of `@NoCSRFRequired` (this one is probably generally something to fix, since it's used for other controllers as well)...

aszlig:
Rebased against master and added throttling.

shreyasajj:
What going on with this PR request?

tacruc:

Issue #92, opened 2019-08-29
Sync with other maps apps (Favorites, Tracks, etc.)


jancborchardt:
We should work on having these apps sync with us:
- [ ] [GNOME Maps](https://wiki.gnome.org/Apps/Maps) might become compatible this summer, ref **Provide APIs to sync with gnome-maps #30**, cc @Sumaid @mlundblad
- [ ] [Marble](https://marble.kde.org/) would require a few changes to be compatible
- [ ] [OsmAnd](https://osmand.net/) would be nice! Nothing like that yet.
- [ ] [Maps.me](https://maps.me/) would be best since they are widely used, but they have a company behind them and already offer their own proprietary favorites backup.

(Carried over from @eneiluj’s notes in https://github.com/nextcloud/maps/issues/56 )

thomas-mc-work:
There is a very actively developed clone of Maps.me called [Maps](https://gitlab.com/axet/omim). Maybe this would be a good candidate for integration? They have already implemented track recording which would play nicely with the sync feature.

> Features over upstream repo:
> 
> - Track Recording (record your tracks)
> - Custom Bookmarks path (helps backup your bookmarks with apps like: Syncthing)
> - Removed ads and binaries (peace and freedom)
> - Screenlock option (prevent app stay turned on while in pocket with lockscreen enabled)

jancborchardt:
@thomas-mc-work good call, yes! Do you have any contacts there or do you want to open an issue in their tracker linking both projects up? :)

thomas-mc-work:
I've raised an issue [here](https://gitlab.com/axet/omim/issues/122), mostly copied from the OsmAnd issue and hopefully phrased as expected.

Where can I find the API documentation? Preferably in the [OpenAPI format](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md).

eneiluj:
@thomas-mc-work Actually there is no Maps Favorites API documentation. It's pretty straightforward though. It's a REST API with 4 routes (listed in `appinfo/routes.php`):

* GET https://mync.org/index.php/apps/maps/api/1.0/favorites => get a JSON list of the favorites
* POST https://mync.org/index.php/apps/maps/api/1.0/favorites => add a favorite, parameters are name (string), lat (float), lng (float), category (string), comment (string), extensions (string)
[...]

Issue #30, opened 2019-03-20
Provide APIs to sync with gnome-maps


Sumaid:
This is part of the ongoing discussion to revive the work in nextcloud maps. I will be proposing a project to sync gnome-maps with nextcloud. Current requirements from the gnome-maps side are
1)Favorites:
This is the most important feature which is required. Having access to your favorite places
where ever you go, is very useful.
Favorites are just one of many possible categories we can have, so API should allow for passing of places in any category. This is very simple to do since we can just have an attribute which stores type of place.
2)Photos with a marked place(Stretch goal):
This will allow users to upload photos with a given place. Usefulness of this feature can be discussed. 

I am still shaping and understanding the proposal so I'll appreciate any ideas and suggestions.

Sumaid:
Data in gnome-maps is stored in a JSON file with the following schema:
{
place:
type:
added:
}
place is an instance of class Place which has attributes:
osm_id, osm_type,name,location,bounding_box,place_type,street_address,street,building,
postal_code,area,town,state,country,county,country_code,continent.

Important attributes of location are longitude, latitude.

eneiluj:
Hi there!

I think we don't want to store osm-related information in NC Maps. A favorite could be defined by:

* name
* type/category
* latitude
* longitude
* note/comment
* date of last modification

The way I see it, the API would make it possible to:

* get all favorites
* get those more recent than a given date
* edit a favorite
* delete
* add

What do you think?

I just checked Gnome-Maps v3.30.1, there's a button to "switch to favorites" but it's disabled. Is the local favorite management feature ready yet?

eneiluj:
Hummm yeah, like you said, a favorite can also be in the middle of nowhere and there's no OSM node to associate with. Also, it's kind of redundant because both OSM nodes and favorites have coordinates. An OSM node could be deduced from a favorite coordinates. It can be ambiguous when there are multiple OSM entities in one place but it doesn't seem problematic.

Anyway favorites don't have to be stored the exact same way in Gnome-Maps and in NC-Maps. If you want to keep a reference to an OSM node in Gnome-Maps, you can. I just think it shouldn't be part of the API to sync with NC-Maps :smile:.

After we get a third opinion I'm ready to implement the API in my rework-eneiluj branch! Maybe @jancborchardt, @rullzer, @juliushaertl or @MorrisJobke have something to add.

juliushaertl:
```
name
type/category
latitude
longitude
note/comment
date of last modification
```

[...]

PR #172, opened 2019-09-27
Add API documentation

thomas-mc-work:
This is a draft to describe the API. I think this will be very valuable for developers of clients.

regarding #92

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1,0 +1,83 @@
+openapi: 3.0.0
+info:
+  title: 'Nextcloud Maps API'
+  version: '1.0'
+  contact:
+    name: Nextcloud Maps Team
+    url: https://github.com/nextcloud/maps/
+externalDocs:
+  description: 'Based on a loose description in this ticket: https://github.com/nextcloud/maps/issues/92'
+  url: 'https://github.com/nextcloud/maps/issues/92'
+servers: 
+  - url: https://{hostname}:{port}/index.php/apps/maps/api/1.0/
+    variables:
+      hostname:
+        default: localhost
+      port:
+        default: '8080'
+paths:
+  /favorites:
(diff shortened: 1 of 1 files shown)


jancborchardt:
Also here: @aszlig @Galbar since @tacruc @eneiluj are a bit busy atm, and you opened pull requests for Maps too (yay! :) it would be lovely if you could review this too. Thanks! :guitar:

openapi.yml
```diff
@@ -0,0 +1,83 @@
...
+          description: Success
+    post:
+      operationId: add-favorites
+      summary: 'Add a favorite'
+      responses:
+        '200':
+          description: Success
+      requestBody:
+        $ref: '#/components/requestBodies/favoriteBody'
+  /favorites/{id}:
```
aszlig: Missing a [`parameters` object](https://swagger.io/specification/#parameterObject) here for `put` and `delete` with `in: path`.

thomas-mc-work: Done.

thomas-mc-work:
How can I add the _DCO_ now?

jancborchardt:
To define the favorites API we should really make sure to loop in the people of other Maps apps to coordinate:

cc @vshcherb @Chumva @nazar-kutz @max-klaus of OsmAnd, and ref their API https://github.com/osmandapp/osmand-api-demo#command-addfavorite
cc @Sumaid @mlundblad from GNOME, and ref **Provide APIs to sync with gnome-maps #30** where we discussed already – @thomas-mc-work did you check that out? :)

Who else?

thomas-mc-work:
> define the favorites API

This PR is not about definition at all. It's only about documenting what is already existing. You are free to change or discuss the API at any time after merging.

PR #173, opened 2019-09-27
Fix Add Device Point error when optional value is empty string.

Galbar:
This happens whith PhoneTrack App.

I found logs of it sending requests with query params as follows: `?lat=1.1&lng=2.2&timestamp=1111111111&user_agent=testDevice&altitude=&battery=95.0&accuracy=36.0`

In the above example the `altitude` is present with a value of empty string (""). This value was passed to the DB sevice, which would fail due to the value not being a number.

The fix just transforms any non-numeric values into null.

### DIFF
--- a/lib/Controller/DevicesApiController.php
+++ b/lib/Controller/DevicesApiController.php
@@ -113,10 +113,10 @@ class DevicesApiController extends ApiController {
      */
     public function addDevicePoint($apiversion, $lat, $lng, $timestamp=null, $user_agent=null, $altitude=null, $battery=null, $accuracy=null) {
         if (is_numeric($lat) and is_numeric($lng)) {
-            $timestamp = normalizeOptionalNumber($timestamp);
-            $altitude = normalizeOptionalNumber($altitude);
-            $battery = normalizeOptionalNumber($battery);
-            $accuracy = normalizeOptionalNumber($accuracy);
+            $timestamp = $this->normalizeOptionalNumber($timestamp);
+            $altitude = $this->normalizeOptionalNumber($altitude);
+            $battery = $this->normalizeOptionalNumber($battery);
+            $accuracy = $this->normalizeOptionalNumber($accuracy);
             $ts = $timestamp;
             if ($timestamp === null) {
                 $ts = (new \DateTime())->getTimestamp();
--- a/lib/Controller/DevicesApiController.php
+++ b/lib/Controller/DevicesApiController.php
@@ -113,6 +113,10 @@ class DevicesApiController extends ApiController {
      */
     public function addDevicePoint($apiversion, $lat, $lng, $timestamp=null, $user_agent=null, $altitude=null, $battery=null, $accuracy=null) {
         if (is_numeric($lat) and is_numeric($lng)) {
+            $timestamp = normalizeOptionalNumber($timestamp);
+            $altitude = normalizeOptionalNumber($altitude);
+            $battery = normalizeOptionalNumber($battery);
+            $accuracy = normalizeOptionalNumber($accuracy);
             $ts = $timestamp;
             if ($timestamp === null) {
                 $ts = (new \DateTime())->getTimestamp();
@@ -169,6 +173,13 @@ class DevicesApiController extends ApiController {
         else {
             return new DataResponse('no such device', 400);
         }
+    }
+
+    private function normalizeOptionalNumber($value) {
+        if (!is_numeric($value)) {
+            return null;
+        }
+        return $value;
     }
 
 }
--- a/tests/Unit/Controller/DevicesApiControllerTest.php
+++ b/tests/Unit/Controller/DevicesApiControllerTest.php
@@ -178,6 +178,14 @@ class DevicesApiControllerTest extends \PHPUnit\Framework\TestCase {
         }
         $this->assertEquals(true, $d2Found);
 
+        // This happens with a request such as /api/1.0/devices?lat=1.1&lng=2.2&timestamp=&user_agent=testDevice&altitude=&battery=&accuracy=
+        $resp = $this->devicesApiController->addDevicePoint('1.0', 1.1, 2.2, '', 'testDevice', '', '', '');
+        $status = $resp->getStatus();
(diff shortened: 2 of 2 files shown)


Galbar:
I tried running the tests but I was unable to get them to run. This is my first time contributing to something nextcloud related. So I'm probably missing something.

I tried running `make test` and got:
```
#/usr/bin/npm run test
No phpunit command available, downloading a copy from the web
mkdir -p /home/Galbar/nextcloud/maps/build/tools
curl -sSL https://phar.phpunit.de/phpunit.phar -o /home/Galbar/nextcloud/maps/build/tools/phpunit.phar
php /home/Galbar/nextcloud/maps/build/tools/phpunit.phar -c phpunit.xml
PHP Warning:  require_once(/home/Galbar/nextcloud/maps/tests/../../../lib/base.php): failed to open stream: No such file or directory in /home/Galbar/nextcloud/maps/tests/bootstrap.php on line 17
PHP Fatal error:  require_once(): Failed opening required '/home/Galbar/nextcloud/maps/tests/../../../lib/base.php' (include_path='.:/usr/share/php') in /home/Galbar/nextcloud/maps/tests/bootstrap.php on line 17
Makefile:171: recipe for target 'test' failed
make: *** [test] Error 255
```

jancborchardt:
And here too: @Galbar @thomas-mc-work since @tacruc @eneiluj are a bit busy atm, and you opened pull requests for Maps too (wohoo! :) it would be very helpful if you could review this too. Thanks! :rocket:

lib/Controller/DevicesApiController.php
```diff
@@ -113,6 +113,10 @@ public function getDevicePoints($id, $pruneBefore=0) {
      */
     public function addDevicePoint($apiversion, $lat, $lng, $timestamp=null, $user_agent=null, $altitude=null, $battery=null, $accuracy=null) {
```
aszlig: The same method is also in `DevicesController`, so maybe fix that as well?

aszlig: Also, I think it's generally a better idea to use parameter types here instead of checking whether a type is numeric or not, for example:

```suggestion
    public function addDevicePoint(int $apiversion, ?float $lat, ?float $lng, int $timestamp=null, string $user_agent=null, float $altitude=null, float $battery=null, float $accuracy=null) {
```

That way it limits the types possible when this method is called. It might also be a good idea to use `declare(strict_types=1);` at the start of the file. However, this might break things for other methods, so it needs to be tested accordingly.

As a side note: There is also [`@param`](https://docs.nextcloud.com/server/17/developer_manual/app/requests/controllers.html#casting-parameters) for use in documentation comments. In theory this should be inferred by type annotations as well, even when using strict types, but I haven't verified whether this really is the case.

lib/Controller/DevicesApiController.php
```diff
@@ -113,6 +113,10 @@ public function getDevicePoints($id, $pruneBefore=0) {
...
         if (is_numeric($lat) and is_numeric($lng)) {
+            $timestamp = normalizeOptionalNumber($timestamp);
+            $altitude = normalizeOptionalNumber($altitude);
+            $battery = normalizeOptionalNumber($battery);
+            $accuracy = normalizeOptionalNumber($accuracy);
```
aszlig: For all of these lines:

```suggestion
            $accuracy = $this->normalizeOptionalNumber($accuracy);
```

Galbar:
I'm unable to run tests in local. Any suggestions?

lib/Controller/DevicesApiController.php
Galbar: is adding the types enough or do we still need the normalization function? Also, what's the meaning of `?` in `?float`? I couldn't find it on google :/

Galbar: Also, I don't understand the difference between `DevicesApiController` and `DevicesController`?

They are _almost_ the same but not. I don't feel comfortable enough with the codebase to try and extract proper abstractions for avoid code duplication.

I can add the same normalization step in `DevicesController` as a quick win, if you prefer it

eneiluj:
Hi @Galbar and @aszlig. Thanks for having a look at Maps!

Let me try to answer the questions here.

@Galbar @aszlig `DevicesController` and `DevicesApiController` are distinct because they don't receive the same requests. The first one manages internal requests from Maps web page so there is a CSRF check. The second one manages requests from clients so there is no CSRF check and CORS headers are sent (allowing non-web clients to perform API requests). Of course if there is a way to put all that in a nicer way, let's do it.

@Galbar About the tests. Phpunit is required to run them. For the moment, they can be launched with (more or less)
```
phpunit --configuration /var/www/html/nextclouphpunit --configuration /PATH/TO/maps/phpunit.xml --coverage-text --color=never
```

@aszlig I agree that strict types should be used and parameters should be typed. I'm like @Galbar, I don't know the exact effect of type restriction on parameters. Feel free to improve anything you want and we'll discuss it.

lib/Controller/DevicesApiController.php
aszlig: The `?typename` means that it is nullable, see https://wiki.php.net/rfc/nullable_types.

aszlig: As for `DevicesController` versus `DevicesApiController`: I don't know the difference either, but just did a quick grep for the method name in question and found it in both. A diff between the two seems that the methods that occur in both are only different by the usage of `@NoCSRFRequired` and `@CORS`, so I think that definitely smells like it could be deduplicated, certainly the `addDevicePoint` function, since their method bodies are the same.

eneiluj: I just explained the difference in the latest comment on this thread.

The code could be factorized indeed but I think we should keep the two distinct controllers to make sure one is getting the CRSF check and the other does not.

eneiluj:
@aszlig Are we all clear on this PR? I'm ok to merge, do you confirm?

jancborchardt:
@eneiluj @tacruc @aszlig @Galbar good to merge then? If so please add your review. :)

tacruc:
Looks good

PR #187, opened 2019-10-13
add Mapbox vector tiles support, not with self hosted servers FTM

eneiluj:
Hey how,

`mapbox-gl-leaflet` has been added and Maps now automatically adds 4 tile servers if Mapbox API key is set:

* streets
* outdoors
* bright
* satellite

This PR targets 0.3 milestone but I think it can be included sooner.

This library can be used to create tile layers getting data from self-hosted OpenMapTiles servers! We "just" need to implement extra tile servers settings and yay!

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -545,7 +545,7 @@
                 }]
             });
             this.map.on('contextmenu', function(e) {
-                if ($(e.originalEvent.target).attr('id') === 'map') {
+                if ($(e.originalEvent.target).attr('id') === 'map' || $(e.originalEvent.target).hasClass('mapboxgl-map')) {
                     that.map.contextmenu.showAt(L.latLng(e.latlng.lat, e.latlng.lng));
                     that.map.clickpopup = true;
                 }
@@ -553,9 +553,8 @@
             this.map.clickpopup = null;
             this.map.leftClickLock = false;
             this.map.on('click', function(e) {
-                if ($(e.originalEvent.target).attr('id') === 'map') {
+                if ($(e.originalEvent.target).attr('id') === 'map' || $(e.originalEvent.target).hasClass('mapboxgl-map')) {
                     if (!that.map.leftClickLock && that.map.clickpopup === null) {
-                        console.log('no popup');
                         searchController.mapLeftClick(e);
                         that.map.clickpopup = true;
                     }
--- a/js/script.js
+++ b/js/script.js
@@ -640,6 +640,7 @@
                     onClick: function(btn, map) {
                         $('.leaflet-control-layers').toggle();
                         $('.easy-button-container').toggle();
+                        that.map.clickpopup = true;
                     }
                 }]
             });
@@ -732,6 +733,7 @@
             }
             $('.leaflet-control-layers').hide();
             $('.easy-button-container').show();
+            this.map.clickpopup = null;
         },
     };
 
--- a/js/script.js
+++ b/js/script.js
@@ -209,25 +209,25 @@
                         maxZoom: 22,
                         attribution: attrib
                     });
-                    mapController.controlLayers.addBaseLayer(mapController.baseLayers['Mapbox vector streets'], 'Mapbox vector streets');
+                    //mapController.controlLayers.addBaseLayer(mapController.baseLayers['Mapbox vector streets'], 'Mapbox vector streets');
 
-                    mapController.baseLayers['Mapbox vector outdoors'] = L.mapboxGL({
+                    mapController.baseLayers['Mapbox outdoors'] = L.mapboxGL({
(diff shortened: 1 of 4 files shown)


tacruc:
Doesn`t work for me. Just shows a monotone color.

eneiluj:
Did you run "make" to get the new JS libs?

eneiluj:
Did you check browser error log? Does it say something mapbox-related?

tacruc:
Sorry I know that this is not alot:
image

eneiluj:
Hmm even if I can't read german, I think I had those when running with "NC 17 beta". It went away with NC 17 stable and wasn't there with NC 16.

Which NC version are you running?

Browser refuses to execute a blob provided by mapbox-gl even if script-src directive authorizes anyone to to it.

tacruc:
I tried the stable17 branche and the v17.0.0 tag both with the same result.

jancborchardt:
Works for me, super nice! :)

But design polishing:

> Maps now automatically adds 4 tile servers if Mapbox API key is set: streets, outdoors, bright, satellite

Things are getting crowded again. ;) When a Mapbox API key is set, we should:
- Replace the default OSM tiles with "Mapbox vector streets", and do not list the OSM tiles at all cause the Mapbox ones look much better and we don’t strain the OSM server.
- Same for the "Mapbox satellite", they should replace the default satellite layer. Except that for the Mapbox satellite layer we also need to overlay the layer with city names as we do currently for satellite as otherwise you don’t know where you are.
- "Mapbox vector outdoors" should replace either the ESRI Topo or the OpenTopoMap, or both.
- We don’t need "Mapbox vector bright"

Generally, keep in mind we should keep stuff simple. :) GNOME Maps for example limits it to 1 great street map and 1 great satellite map, same for Google Maps.

eneiluj:
@tacruc I guess you should try with [the stable release archive](https://download.nextcloud.com/server/releases/nextcloud-17.0.0.zip).

eneiluj:
@jancborchardt Yes I agree. I'll do it when I find time again :wink:.

tacruc:
Nice. Got it working with the tarball.
I have the feeling the satellite gets artificially blurred if you zoom in?

jancborchardt:
> @jancborchardt Yes I agree. I'll do it when I find time again wink.

Cool! :)

> It does not work. I think the overlay is displayed behind the mapbox-gl "canvas".

Right, apparently the one we want is "[Mapbox satellite streets](https://www.mapbox.com/maps/satellite/)" instead of "Mapbox satellite". Is that possible?

eneiluj:
> "Mapbox satellite streets" instead of "Mapbox satellite". Is that possible?

It's more than possible, it's done :wink:

So here are the news:

* If a Mapbox API key is set, the layers that are accessible with the buttons (main layers: OSM and ESRI Aerial) are replaced by Mapbox streets and Mapbox satellite streets. Dark, OpenTopoMap and ESRI topo are removed. Mapbox Dark and Mapbox outdoors are added to the layer list.
* When left-clicking the map in order to close layer selector, "place" popup was showed. It's now fixed.

If there is no Mapbox API key, OpenTopoMap and ESRI topo are a bit redundant but as neither of them is as good as Mapbox outdoors, I think we can keep them both to let user choose his/her favourite one... What do you think?

It is so much nicer now! Ultra yay! :tada:

jancborchardt:
Aaaawesome! :tada: 

> If there is no Mapbox API key, OpenTopoMap and ESRI topo are a bit redundant but as neither of them is as good as Mapbox outdoors, I think we can keep them both to let user choose his/her favourite one... What do you think?

Yeah … I would actually say that ESRI Topo looks much better and we can remove OpenTopoMap, but maybe @tacruc as an actual hiker has a different view here? ;)

Also I’m wondering – can we keep Stamen Watercolor (incl the streets layer), or is there a Mapbox version of it? It’s just such a cool little easter egg imo, even though it has some issues like not displaying Chinese, Japanese and Korean characters (which might be fixable using some Mapbox layer?).

eneiluj:
> can we keep Stamen Watercolor (incl the streets layer), or is there a Mapbox version of it?

After a quick search I think there is no out-of-the-box watercolor available from Mapbox. We can keep it as it's nice and not so disturbing.

> ESRI Topo looks much better and we can remove OpenTopoMap

OpenTopoMap is less nice but displays much more information than ESRI Topo.

Sorry to add an unrelated commit here but I found a small bug which does not deserve a PR:
When Maps is accessed with untrusted SSL certificate, browser refuses to provide location so location control fails but stays activated, leading to an error message on each page load. It's now solved.

jancborchardt:
Heh, a second after pressing the merge button I was actually wondering if this should have gone in already. :D But I see no issue?

Issue #169, opened 2019-09-19
Display contacts only from enabled addressbooks


efelon:
Nextcloud 16.0.4
Maps 0.1.2

Maps App should only displays contacts from enabled address books. At the moment it show also disabled ones. 
This helps to "bulk-hide" contacts now, since there is no additional filtering in the app yet, but also in general imho.

Best regards
efelon

jancborchardt:
Yup, this should basically simply mirror what is enabled or not in the Contacts app. No additional interface options needed.

eneiluj:
Yep I agree. I'd like to only show contacts from enabled address books.

@skjnldsv Could you give me a hint about how to know if an address book is enabled or not in Contacts app from another app? I get the address book objects from NC contacts manager and also directly from the cardDav backend but I didn't find out how to know which address book is enabled.

skjnldsv:
There is a dedicated prop for it.
No idea how you can get it in php, maybe it's straightforward? cc @georgehrke

georgehrke:
Looking at https://github.com/nextcloud/server/blob/master/apps/dav/lib/CardDAV/AddressBookImpl.php I would say that it's not exposed at the moment.

eneiluj:
So how does the Contacts app know which one is enabled?

georgehrke:
It uses its own CardDAV property which is stored inside oc_properties:

![19E96B59-2EA8-4562-BFC2-76DDD25F0EA9](https://user-images.githubusercontent.com/1250540/66760384-46b6ab80-eea2-11e9-9de1-2bac18e9b58c.png)

eneiluj:
Yay thank you.

PR #192, opened 2019-10-14
refs #169 don't get contacts from disabled address books

eneiluj:
As mentioned in #169, contacts from disabled address books were showed on the map and suggested when placing contacts. Address book state is now verified from the database.

### DIFF
--- a/lib/Controller/ContactsController.php
+++ b/lib/Controller/ContactsController.php
@@ -58,7 +58,20 @@ class ContactsController extends Controller {
         $addressBooks = $this->contactsManager->getUserAddressBooks();
         $result = [];
         $userid = trim($this->userId);
+        // get addressbook ids
+        $bookids = [];
         foreach ($contacts as $c) {
+            if (!in_array($c['addressbook-key'], $bookids)) {
+                array_push($bookids, $c['addressbook-key']);
+            }
+        }
+        // determine if addressbooks are enabled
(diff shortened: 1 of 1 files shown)


jancborchardt:
@efelon can you check if this fixes the issue you reported at https://github.com/nextcloud/maps/issues/169 ? :)

georgehrke:
Please note that you are ever only supposed to query ( or insert / update / delete ) tables that your app also created.

`oc_properties` is a private database table created by the `dav` app and may change at any time without prior notice. 

Meaning, right now there is no officially supported API to get the enabled status of an addressbook in PHP. I created this issue and will try to look into it. https://github.com/nextcloud/server/issues/17537

eneiluj:
@georgehrke Ok thanks for the feedback.

So we shouldn't merge that for now.

Now let's keep this sleeping until there is an official way to get address book states.

efelon:
@jancborchardt in that case there is no need for me to test. I'm aware and will wait for a final solution.

Thanks anyway everyone for looking into it.

JoomlaStats:
It seems to be a trivial feature request to hide disabled contacts from showing up in the map. For example if someone has an 'archive' address book. 
Meanwhile it should have been possible to get an 'API' agreement with the coders from the contacts app? Otherwise, I would hope that meanwhile we're stop waiting and implement the enable/disable state based on the oc_properties, taking the dependency for granted, while waiting for the API-agreement. As it's taking the functionality down big time.

PR #195, opened 2019-10-15
Improve design of Contacts tooltip and popup

jancborchardt:
Just to finish off for the day, a still kind-of work-in-progress version. Feel free to test @nextcloud/maps

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -123,6 +123,7 @@
 }
 .leaflet-container {
     background: var(--color-main-background);
+    font-family: var(--font-face);
 }
 
 .leaflet-control-layers-base {
@@ -279,6 +280,7 @@ tr.selected td {
 }
 .leaflet-popup-content-wrapper, .leaflet-popup-tip {
     background-color: var(--color-main-background) !important;
+    box-shadow: none;
 }
 .leaflet-bar {
     background-color: var(--color-main-background);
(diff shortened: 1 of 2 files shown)


eneiluj:
Much nicer!

Why did you hide the delete icon and the right click popup? I mean, do you think it brings confusion somehow?

Also, I think it's better to comment the inside of `onContactRightClick` method rather than the whole method. Now there is an error because the method is not found.

jancborchardt:
> Why did you hide the delete icon and the right click popup? I mean, do you think it brings confusion somehow?

Yes, I think we should use right-click as little as possible. In this case it brings up a single-entry-list popup which you do not see in any other case. It would totally be enough for the "Delete address" button to just be in the popup when you click.
(Actually I had the case where I had the popup open and then right-clicked, and the popup opeend over another.)

> Also, I think it's better to comment the inside of onContactRightClick method rather than the whole method. Now there is an error because the method is not found.

Right, I’m a noob, will fix that. :D

eneiluj:
> It would totally be enough for the "Delete address" button to just be in the popup when you click.

I think so too. That's why I wondered why you removed it from the popup too.

jancborchardt:
> > It would totally be enough for the "Delete address" button to just be in the popup when you click.
> 
> I think so too. That's why I wondered why you removed it from the popup too.

Yeah, sorry :D I needed to clean up stuff and start simple to see where stuff should go first. :) As said it’s work in progress.

Other feedback though? One thing I’m thinking is that actually:
- It might be nicer looking to put contacts just in a circle without a pin, and the circle is centered on their address. (Possible concern: If you have just a city name set, they could overlay the city name on the map.)
- It’s probably better after all to show the extra tooltip and popover _above_ the avatar marker, but significantly so it does not overlap the avatar marker at all – unlike we had it before. Main reason was that during use, when I hover it the actual click area (the marker) is overlapped, which is then weird and I don’t necessarily know I can click it.

Will work on that more here. :)

These GNOME Maps designs are quite good, from https://wiki.gnome.org/Design/Apps/Maps

eneiluj:
> It might be nicer looking to put contacts just in a circle without a pin

Then it would not be consistent with the other markers... I don't know.

> and the circle is centered on their address (Possible concern: If you have just a city name set, they could overlay the city name on the map.)

Well we're not sure the city name is precisely on city coordinates returned by Nominatim.

> It’s probably better after all to show the extra tooltip and popover above the avatar marker,

I totally agree. I was weirder to me to have the tooltip hiding the marker. It's better to put it away and avoid the overlap.

No more idea for the moment.

eneiluj:
After a second thought, it's probably better to remove the pin from all markers (device, track, contact) and center them on the element position like it is already done for favorites. Then it would feel more natural to see a tooltip or a popup appear on top of them with a pin.

Otherwise there is kind of a pin repetition.

I thought that without a pin there would be an ambiguity between a marker and a cluster of markers but the circle with the number of clustered elements if good enough to make the distinction.

Other point to consider, with a pin pointing to the position, the actual location is not hidden by the marker. Centering a pinless marker on the location hides it. Not a big deal...

eneiluj:
Another thing, the aspect of the avatar is preserved in the marker but not in the popup.

jancborchardt:
> Other point to consider, with a pin pointing to the position, the actual location is not hidden by the marker. Centering a pinless marker on the location hides it. Not a big deal...

And that was the concern I mentioned above, hiding the city name e.g.

After looking more at Google Maps and Apple Maps, I would also keep it like we have it now as they do it similarly. We can look at specific cases like when search brings up many results – then it makes sense to just use pins for the main results and points for others, also like Google Maps does.

tacruc:
I just tried and liked it :+1:

eneiluj:
I just found the `object-fit: cover` css property which is perfect for preserving aspect of contact popup avatar (img tag).

For marker thumbnails, `background-size: cover` is used.

jancborchardt:
@eneiluj good point! I think contact avatars are actually always saved as square, at least by Nextcloud Contacts, right @skjnldsv?

tacruc:
@jancborchardt: I think we cannot relay on this, 50% of at least my contacts have non square images.

skjnldsv:
@jancborchardt no, we do not edit. We save the photo as given by the user :)

eneiluj:
Yep I mentioned `object-fit: cover` because most of my contact's pictures are not squared shaped. @jancborchardt If yours are all square, maybe the tool you use to crop them before setting them makes sure they are square.

tacruc:
@jancborchardt  Whats the status here?

PR #196, opened 2019-10-16
fix bug in names making dark and topographic inaccessible

eneiluj:
I made a mistake when reviewing #187. As it's implemented, layer visible names have to be the same as layer dict keys. Dark and topographic didn't work anymore.

It's working now.

I don't have my GPG key on the computer I'm using right now so my commit is not signed...


eneiluj:
signed, pushed and merged. thanks

PR #227, opened 2019-12-22
use standalone Viewer app when clicking on photo

### LINKED ISSUE #93: Open Viewer when clicking on photo
When clicking on a photo, right now it opens a new tab with the files view, which breaks the flow.

Much better would be to simply open the [Viewer](https://github.com/nextcloud/viewer/) with the photo. Navigation left-right in the viewer should probably work just like in the folder, to go to the next newer or previous older picture. (In the case of a track this will also correspond to where you were first etc.)

eneiluj:
As mentioned in https://github.com/nextcloud/viewer/issues/157 it's now possible to use Viewer app in other apps as it's not strongly depending on Files app anymore.

So now, if Viewer app is recent enough, it's used directly in Maps when a photo is clicked. Otherwise a new tab to Gallery app is opened.

If you wanna try it, install a recent version of the Viewer app. I used `v18.0.0beta3`.

closes #93

### DIFF
--- a/js/photosController.js
+++ b/js/photosController.js
@@ -110,10 +110,16 @@ PhotosController.prototype = {
         var _app = this;
         return function(evt) {
             var marker = evt.layer;
-            var galleryUrl = OC.generateUrl('/apps/gallery/#'+encodeURIComponent(marker.data.path.replace(/^\//, '')));
-            var win = window.open(galleryUrl, '_blank');
-            if (win) {
-                win.focus();
+            // use Viewer app if available and recent enough to provide standalone viewer
+            if (OCA.Viewer && OCA.Viewer.open) {
+                OCA.Viewer.open(marker.data.path);
+            }
+            else {
+                var galleryUrl = OC.generateUrl('/apps/gallery/#'+encodeURIComponent(marker.data.path.replace(/^\//, '')));
+                var win = window.open(galleryUrl, '_blank');
+                if (win) {
+                    win.focus();
+                }
             }
         };
     },
--- a/js/photosController.js
+++ b/js/photosController.js
@@ -13,6 +13,7 @@ function PhotosController (optionsController, timeFilterController) {
     this.timeFilterEnd = Date.now();
 
     this.movingPhotoPath = null;
+    this.isPhotosInstalled = OCP.InitialState.loadState('maps', 'photos');
 }
 
 PhotosController.prototype = {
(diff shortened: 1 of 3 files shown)


js/photosController.js
```diff
@@ -110,10 +110,16 @@ PhotosController.prototype = {
...
-            var galleryUrl = OC.generateUrl('/apps/gallery/#'+encodeURIComponent(marker.data.path.replace(/^\//, '')));
-            var win = window.open(galleryUrl, '_blank');
-            if (win) {
-                win.focus();
+            // use Viewer app if available and recent enough to provide standalone viewer
+            if (OCA.Viewer && OCA.Viewer.open) {
+                OCA.Viewer.open(marker.data.path);
+            }
+            else {
+                var galleryUrl = OC.generateUrl('/apps/gallery/#'+encodeURIComponent(marker.data.path.replace(/^\//, '')));
```
skjnldsv: Careful, gallery will not be shipped nor be compatible with 18 :)

eneiluj: Ow alright. So any idea about what to do if Viewer is not there or not recent enough?

skjnldsv: replacement app is named `photos`
https://github.com/nextcloud/photos

tacruc: Photos is using the viewer app for the sildeshow, or?
And the gallery is the fallback for NC17 and earlier. So we just need to add a small check if the gallery app is enabled, or?

tacruc:
lgtm, as soon @skjnldsv is happy we should merge it.

js/photosController.js
eneiluj: Is it possible to know if an app is enabled from the frontend point of view without doing an ajax call? I didn't find it with OC/OCA/OCP.

skjnldsv: > Photos is using the viewer app for the sildeshow, or?

yes

> Is it possible to know if an app is enabled from the frontend point of view without doing an ajax call? I didn't find it with OC/OCA/OCP.

You can do that with an initialstate from php and use it in your app.
https://github.com/nextcloud/photos/blob/6f808ea8ef4d30cab46bc2227e47132142f57176/lib/Controller/PageController.php#L74
https://github.com/nextcloud/photos/blob/6f808ea8ef4d30cab46bc2227e47132142f57176/src/services/IsMapsInstalled.js#L25
https://github.com/nextcloud/photos/blob/6f808ea8ef4d30cab46bc2227e47132142f57176/src/router/index.js#L30

eneiluj: Thanks for the hint. Done and pushed.

The Viewer app is not visible in application settings in NC 18.0.0beta3 though it's installed, functional and OCA.Viewer available. I wanted to disable it to check the generated links to Photos app. Anyway I tried them manually and it works fine.

In Photos, there is no more route pointing directly to a picture like in Gallery app. It was handy for this use case. I guess it's intentional.

So now, if standalone Viewer is not available and Photos apps is enabled, Maps app opens a tab to Photos app in the album view containing the clicked picture.

tacruc: Works, but the photos app is not working, if the viewer app is not available.

tacruc: Tested works fine

js/photosController.js
```diff
@@ -13,6 +13,7 @@ function PhotosController (optionsController, timeFilterController) {
     this.timeFilterEnd = Date.now();
 
     this.movingPhotoPath = null;
+    this.isPhotosInstalled = OCP.InitialState.loadState('maps', 'photos');
```
skjnldsv: At some point it would be nice to have packages and compile in maps ;)

eneiluj: Yep, agreed. I don't have any experience with that. Is there anything more to do that following https://docs.nextcloud.com/server/17/developer_manual/app/npm.html instructions about webpack?

skjnldsv: It depends to be fair.
I'm not familiar with the maps app, but you might benefit from switching ot vue and use the nextcloud components :)
With this, you can have a look at the nc app tutorial or the vueexample app

PR #241, opened 2020-01-18
Fix issue 239

tacruc:


tacruc:
If you have mapbox enabled, but disable webgl in your browser you will see an empty map. With this PR there should be the non mapbox map loaded.
New:
image

Old:
image

jancborchardt:
@djohannes can you review if this fixes the issue you reported at https://github.com/nextcloud/maps/issues/250 ? :)

Issue #242, opened 2020-01-18
make sed command fails on macOS


jamesstout:
```bash
james@Jamess-iMac: ~/Projects/nextcloud/server/apps/maps on master
$ make
...
sed -i 's/L\.Browser\.touch/L.Browser.mobile/g' node_modules/leaflet.elevation/dist/Leaflet.Elevation-0.0.2.min.js
sed: 1: "node_modules/leaflet.el ...": extra characters at the end of n command
make[1]: *** [npm] Error 1
make: *** [build] Error 2
```
A portable (not POSIX though, `sed -i` is not POSIX-compliant) fix is:
```bash
sed -i.bak 's/L\.Browser\.touch/L.Browser.mobile/g' \
node_modules/leaflet.elevation/dist/Leaflet.Elevation-0.0.2.min.js && rm \
node_modules/leaflet.elevation/dist/Leaflet.Elevation-0.0.2.min.js.bak
```

jamesstout:
Submitted a PR: #243 for review.

PR #243, opened 2020-01-18
fix Makefile sed on macOS

jamesstout:
Addresses issue #242.

Tested on macOS 10.13.6 and Ubuntu 18.04

Signed-off-by: James Stout <stoutyhk@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,8 +87,8 @@ endif
 .PHONY: npm
 npm:
 	$(npm) install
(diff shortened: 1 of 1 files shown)


jancborchardt:
@jamesstout thanks a lot for your contribution!

Can someone on macOS test? :) cc @georgehrke

jancborchardt:
Hmm, does this maybe break it on non-macOS? cc @tacruc @paulschwoerer @eneiluj

v1r0x:
The command still works on my ubuntu and should work on (all?) systems supporting `-i`. But I don't get why macOS doesn't like `sed -i`, but `sed -i.bak` is fine :thinking:

PR #299, opened 2020-02-16
Use global jQuery for now

gary-kim:
As it turns out, we still need to use the global jQuery else we get CSRF errors. :cry: 

Especially since the global jQuery is being deprecated (nextcloud/server#19386), we need to move away from using it ASAP but this is a quick fix so we can get `0.1.5` out and fix the unable to unpack issue. Let's switch to using `@nextcloud/axios` before the release of Nextcloud 19.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4241,11 +4241,6 @@
       "integrity": "sha1-R+Y/evVa+m+S4VAOaQ64uFKcCZo=",
       "dev": true
     },
-    "jquery": {
-      "version": "3.4.1",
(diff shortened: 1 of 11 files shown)


ChristophWurst:
Alternatively you can copy https://github.com/nextcloud/server/blob/master/core/src/jquery/requesttoken.js and still use your own jquery

tacruc:
@gary-kim what do you think?

gary-kim:
There's a couple other features from core jQuery that's being used so that needs to be changed first so I think it'd be better to postpone that for a little bit.
```
TypeError: $.ui is undefined script.js:1530:13
```

Issue #307, opened 2020-02-21
Favorites not showing on maps


RadoslawDryzner:
Hello,

I have noticed a bug in the latest version when running the maps app within Nextcloud. Simply said, the favorites added on the map do not show on the map. If you inspect the responses to requests done by the browser, you can see the favorites are actually returned with their ID and coordinates, but simply do not appear on the map. The favorites icon on the left is always set to "loading".

The setup I am running is the following. It should be very easy to reproduce.

* I am running latest Nextcloud image from docker. I believe the Nextcloud version there is 18.0.1.
* It happens from a completely fresh install using either SQLite or PostgreSQL databases.
* I install Maps from the App Store. It is on version 0.1.5 I think.
* I am using the latest version of Firefox on Linux to access the interface.

Steps to reproduce:
1. Pull nextcloud image from docker hub.
```docker pull nextcloud```
2. Run the docker image in a new container.
```docker run -d -p 8080:80 nextcloud```
3. Go with the first run setup choosing SQLite.
4. Access Nextcloud on `localhost:8080`
5. Install Maps with the app managment tool within Nextcloud.
6. Open Maps app.
7. Add a favorite.

The whole setup shouldn't take longer than 5 minutes to reproduce.

I am quite new in Netxcloud so I am not familiar with which logs could be relevant here. If you tell me which ones are important, I am happy to provide them.

r3pek:
Same as #305

doc75:
@r3pek, as stated in #305 I was not able to validate the fix. Not sure if this is a mistake in my test procedure or not.

RadoslawDryzner:
Sorry I wasn't aware of the other issue. How quick do changes on master usually move into release here?

Bergum:
After updating to 0.1.5 i have ony:
My favorites are just spinning,
One picture on map.
No tracks are loading (Just spinning) (but are visible in right menu, but all hidden/displayed and color are reset to default. Cant get them on the map.


Tracks where working before the update.

NK Version 18.01 upgraded from 18.0
Maps Ver 0.1.5

eneiluj:
@Bergum @RadoslawDryzner @r3pek Could you check the browser console error log and show complete error lines?

I'm quite surprised the fix does not work. Let's find out and fix it!

Bergum:
Could you check the browser console error log and show complete error lines?

How do that?

eneiluj:
[...]

Issue #305, opened 2020-02-19
v0.1.5 does not show favorites anymore


doc75:
Environment: docker official image (fpm-alpine)
Nextcloud version: 16.0.7 & 16.0.8

I was on maps v0.1.2 and tried to upgrade to maps v0.1.5.
After doing so I cannot see anymore my favorites.

To try to figure out which version is breaking things, I did the following:
1. deactivate maps from Nextcloud UI
2. suppress maps from Nextcloud UI
3. go in custom_apps directory
4. download version 0.1.2
5. extract version
6. change ownership: `chown -R www-data:www-data maps`
-> Favorites can be seen as before.

I did the same 6 steps for version 0.1.3
-> Favorites can be seen as before.

I did the same 6 steps for version 0.1.4
-> Favorites can be seen as before.

I did the same 6 steps for version 0.1.5
-> Favorites **cannot** be seen as before.

Contacts in all version can still be seen.

I cannot find anything in the log that seems related to this issue.

Any idea on what I can do to try to get more information ?

eneiluj:
Thanks for the bug report. I think it's a side effect of merging frontend code with webpack. It's fine now with #306. If you feel like trying, it's available in master branch now.

doc75:
I Here is what I tried, but still having the same issue:
```sh
apk add nodejs npm make
wget https://github.com/nextcloud/maps/archive/master.zip
unzip  master.zip
mv maps-master/ maps
cd maps/
make
cd ..
chown -R www-data:www-data maps/
```
Then I activate the maps application in Nextcloud.
I still have the same issue.

I can see the modification you made through #306, but in the Web console of firefox I can see the same error: `ReferenceError: assignment to undeclared variable cat`

Did I miss something to test your modification properly ? Sorry but it is the first time I am testing a master version of a Nextcloud app.

doc75:
@eneiluj could we reopen it while fix is not validated ?

eneiluj:
Yes we can.

Are you sure it's the same error. I mean is it still on the same line of same file?

eneiluj:
Let's close this issue if we find out it's exactly the same as #307. Not sure yet.

doc75:
@eneiluj, thanks for the advice, I cleared the cache (Ctrl+F5 does not seem to do anything on firefox), then it leads to another error.

As it is mentionned in script.js which is having very big line, the debug is not easy (sorry for the silly question, but is there a way to work on the `src` version to debug easily ?).

[...]

PR #309, opened 2020-02-22
fix another missing var in tracksController.js

eneiluj:
refs #305 #307 

This was preventing tracksController to work.

@doc75 After a few tests I didn't find other places where this kind of problem occurs. Of course a second round of tests is very welcome.

Once we are good on this, maybe we should think about publishing v0.1.6. What do you think @tacruc?

### DIFF
--- a/src/tracksController.js
+++ b/src/tracksController.js
@@ -547,6 +547,7 @@ TracksController.prototype = {
         });
 
         var trks = gpxx.find('trk');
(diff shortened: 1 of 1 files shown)


tacruc:
Yeah I guess we should publish everything which fixes bugs fastly.

doc75:
@eneiluj I tried the master version and it seems to be OK for Favorites, Tracks and Contacts (other points not tested as i do not have associated data).

New release is welcome.

Thanks for this great app.

PR #292, opened 2020-02-11
Use Webpack

gary-kim:
Closes #252 

Being a bit conservative with the milestone for now. :smile: 

There's quite a lot more that could be done for cleanup but I think it'd be better to put that in a follow-up PR.

EDIT: Someone who has access to settings for the repo: could you enable the `projects` feature? I'd like to make a project for migration to Vue.
EDIT 3: Once this is done, PR incoming for migrating admin settings to Vue.
EDIT 4: 
@eneiluj @jancborchardt `package-lock.json` and `composer.lock` being in the `.gitignore` file caused me some confusion. I just read #44.
I just removed those lines from `.gitignore` and committed those files in this PR.
Just for future reference, `package-lock.json` and `composer.lock` should be committed into source control: [package-lock.json documentation](https://docs.npmjs.com/files/package-lock.json) && [composer.lock documentation](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control) :smile:. (I don't mean that with any negative intent)

PR #310, opened 2020-02-22
update composer and npm metadata files

eneiluj:
Those files are local and don't need to be tracked by the repo.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "^7.8.3"
+        "@babel/highlight": "7.8.3"
       }
     },
     "@babel/compat-data": {
@@ -19,9 +19,9 @@
       "integrity": "sha512-jWYUqQX/ObOhG1UiEkbH5SANsE/8oKXiQWjj7p7xgj9Zmnt//aUvyz4dBkK0HNsS8/cbyC5NmmH87VekW+mXFg==",
       "dev": true,
       "requires": {
-        "browserslist": "^4.8.5",
-        "invariant": "^2.2.4",
-        "semver": "^5.5.0"
+        "browserslist": "4.8.6",
+        "invariant": "2.2.4",
+        "semver": "5.7.1"
       }
     },
     "@babel/core": {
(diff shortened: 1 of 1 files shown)


tacruc:
https://github.com/nextcloud/maps/pull/292#issuecomment-584630014

eneiluj:
Ok thanks for pointing to the comment.

Still it's a bit annoying since it seems only required to avoid the automatic npm "test" to fail. I would prefer to get rid of those and to make changes in the proper place to generate them before calling `npm ci`. What do you think?

tacruc:
I have no preference, but maybe @gary-kim or @ChristophWurst pointing this out at the first time.

gary-kim:
The point of dependency lock files are to ensure everyone has the exact same set of dependencies installed, no matter their environment. They are meant to be in the repo. They help make sure there are never any bugs that arise from having a different version of a dependency, speed up installing dependencies, and also help with security by storing the hash of all dependencies.

[package-lock.json documentation](https://docs.npmjs.com/files/package-lock.json) && [composer.lock documentation](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control)

There's also some documentation about the reasons for lockfiles in npm [here](https://docs.npmjs.com/files/package-locks)

Whenever updates are required for dependencies for Javascript, npm, or in our case, dependabot, will automatically update the package-lock.json file as well. The same will happen for composer dependencies as well.

eneiluj:
@gary-kim Ok thanks for the explanation. So we should not pay attention to the fact we will commit them often? We just commit them and that's it?

gary-kim:
Normal changes should not be updating the lockfiles unless they involve changing dependencies. The fact that the app store template includes `composer update` as part of the setup process is strange. @ChristophWurst do you know why that's there?
https://github.com/nextcloud/maps/blob/c5e1f99cc5c5038d872abfb2dbb7e97b93d86a8f/Makefile#L79-L83

eneiluj:
Thanks a lot for looking into it!

It might just be something to fix in the Makefile indeed.

ChristophWurst:
> @ChristophWurst do you know why that's there?

Good catch. That doesn't seem right. Please fix the Makefile and also submit that for the template.

eneiluj:
Thanks for #314.

So I reverted everything on this branch. Rebased on master and ran `make` which updated `package-lock.json`. All clear?

gary-kim:
That's strange. What version of npm do you have installed? My system doesn't make that change.

eneiluj:
Node version is v10.15.2 and npm is 5.8.0.

After upgrading npm to 6.13.7, `package-lock.json` is not changed anymore by running make.

Should we close this or is there anything left to do?

Issue #290, opened 2020-02-07
re-enable contextmenu "place photo folder" that was removed in the past


wronny:
Hello, 
I'm like to use maps to add pictures. Unfortunately contextmenu "place photo" only allow to add single photo or several using ctrl+mouseclick. However this is not very comfortable to select many photos. While investigating on adding photos by folder I stumbled over "place photo folder" in script.js. Unfortunately this has been removed / commented in the past:

![grafik](https://user-images.githubusercontent.com/3709024/74073263-014a5b00-4a0a-11ea-981f-a5fcc340a0a3.png)

I tested using my nc18 maps to re-add this contextmenu item and surprisingly "place photo folder" was working without any further modification. 
Is there any reason that I might have missed that "place photo folder" should be kept disabled? Maybe it is just still disabled just because it has been forgotten to re-enable it again?
Maybe someone else likes to verify my observation?
Many thanks
Wronny

wronny:
addendum: If I use "place photo folder" re-enabled in js/script.js it takes about a minute until new added photos are shown on map. Not sure why this is different to (single) "place photo" that does show the added photo immediately.

jancborchardt:
@wronny would you like to open a pull request to add this back? And maybe you can look into the issue of the delay until photos are shown on the map? :)

wronny:
fixed by https://github.com/nextcloud/maps/pull/316
Many thanks to @eneiluj

PR #316, opened 2020-02-25
Fix place photos and bring back place dirs

eneiluj:
refs #290

Hey! I'm kind of back.

I think, like @wronny, that photo folder placement is a nice feature.

Here is what I suggest to make photos and directories placement working and instantaneous. It was only working when the photo already was in the DB which is not the case (most likely my mistake :grin: or a recent change that broke that).

Feel free to tell me it should have been done in 2 separate PRs :wink:.

### DIFF
--- a/lib/Service/PhotofilesService.php
+++ b/lib/Service/PhotofilesService.php
@@ -162,7 +162,6 @@ class PhotofilesService {
             $this->photoMapper->deleteByFileIdUserId($fileId, $userId);
         }
     }
-    
 
     public function deleteByFolder(Node $folder) {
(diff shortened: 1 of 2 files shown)


wronny:
@eneiluj , working perfect. 
BTW: Looks like in maps 0.1.5 also "place photos" was delayed but your change in PhotofilesService.php fixed it for both "place photos" as well as for "place photo folder".
Many thanks.

jancborchardt:
@eneiluj you closed this, can the branch be removed? :)

eneiluj:
Yes it can :grin:

PR #322, opened 2020-02-29
fix delay when placing photos/folders

eneiluj:
refs #290

Here is what I suggest to make photos and directories placement working and instantaneous. It was only working when the photo already was in the DB which is not the case.


jancborchardt:
@wronny could you review this, as it refers an issue you have been involved in? :) Thanks a lot!

PR #331, opened 2020-03-08
Fix and polish for 0.1.6

eneiluj:
I was about to publish a new release but I found a lot of small things to here and there.

I think we're very close to v0.1.6 with that bunch of fixes and small polishing.

This PR includes:
* enlarge some very small tooltips
* fix style of favorite category which have space in their names (same for contact groups)
* fix cursor on mapboxgl map (when loading, moving things, grabbing map etc...)
* fix all popup behaviour:
    * close them when not needed anymore (delete favorite with right-click for example)
    * properly manage interference with left click on map (map.clickpopup flag) on many events (closing popups with close button for example)
    * adjust position of some misplaced ones
* harmonize "add favorite" with "move photo" and "move favorite": display message and allow canceling by pressing ESC
* fix zoom on small tracks: maxZoom=17
* fix bug when updating device positions while having some disabled ones
* move L.Elevation patch before webpack build to make sure it's effective
* add "dev" and "npm-dev" Makefile targets that perform a faster webpack dev build

Feel free to point remaining important things to do before v0.1.6 like blocking bugs or interface glitches.


tacruc:
bump the version

Bergum:
Should there be a difference between unstable and stable version?
I did update cause nextcloud told me a new version was online. 
It's only on the webpage it's written that it can be unstable, and there is no choise to only install stable ones..

Issue #293, opened 2020-02-12
When clicking on a group of photo : Choose weather to see all photos or zoom on the zone


loupdavid:
At the moment, when I click on a group of photo on the map, it will either zoom on the zone, or open a slideshow depending on the number of grouped photos or their geographical distances (?)

It would be an appreciated feature to be able to see a slideshow of a huge number of photo, over a whole country or city. It might be an option on right click to show the photos and keep the current behavior for the left click.

dennorske:
I do not necessarily agree that this would be as nice, although I understand what you mean. I am using the current design where it zooms in if there are sub-groups of pictures available. 
Another way of attacking this would be to search for City / Country of origin (by having an API in the backend to tell maps which coordinates is where, not sure of OSM can tackle that). 

1. Search for City / Country you want pictures from
2. Maps zoom appropriately to the area while scanning
3. Done with the scanning, maps generates a quick index for the results and let you browse it in slideshow view

I think that would avoid changing the current behavior too much and confuse people :)

wronny:
I like the approach suggested by @loupdavid to be more flexible in choosing whether to zoom into a photo cluster or to view the whole current photo cluster in photo viewer. Therefore I created the above mentioned forked commit to create a clustercontextmenu for photo clusters. "Zoom in" does zoomToBounds like a click and "View all photos" will open the photos of the whole cluster in photo viewer:
![2020-03-10 23_23_33-Karten - Nextcloud](https://user-images.githubusercontent.com/3709024/76370609-b2108680-6337-11ea-89c9-121c992c2b87.jpg)
Maybe someone like to review this. I'm open for any comment / feedback. 
What about @eneiluj or @jancborchardt ? Like to have a view on this?

tacruc:
@wronny, can you create a PR to master?

tacruc:
Thinking oft this would, I don't See a good reason why is should be nice to have a slideshow of more than 10 elements in a random order.
I'm not convinced that this is imporant, can you name one new usecase?

But remember the order of the elements is uncontrolled right now.

loupdavid:
Here's my usecase, it's based on my personal experience with maps, so I don't know how many other users may need this too.
I've made journey where I visited multiple cities in the same country and took lots of photos. When watching the photos, I'd like to have a slideshow for _cityA_, then one for _cityB_ and one for _cityC_, ... . But as I took many photos in each cities, I can't have all photos of a specific city in **one slideshow**.
[...]

PR #332, opened 2020-03-11
photoClusterContextmenu

wronny:
I like the approach suggested in #293 by @loupdavid to be more flexible in choosing whether to zoom into a photo cluster or to view the whole current photo cluster in photo viewer.

See #293 for some pros and cons :-)

### DIFF
--- a/src/photosController.js
+++ b/src/photosController.js
@@ -52,6 +52,8 @@ PhotosController.prototype = {
                 }
             }
         });
+        //right click on photo cluster:
+        this.photoLayer.on('clustercontextmenu', this.photoClusterContextmenu);
         // click on photo menu entry
         $('body').on('click', '#navigation-photos > a', function(e) {
             that.toggleLayer();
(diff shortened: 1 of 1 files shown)


src/photosController.js
```diff
@@ -238,6 +264,37 @@ PhotosController.prototype = {
...
+        var zoomText = t('maps', 'Zoom in');
+        var res =
+            '<ul layerId="' + layerId + '">' +
+            '   <li>' +
+            '       <button class="icon-play viewphotos">' +
+            '           <span>' + viewText + '</span>' +
+            '       </button>' +
+            '   </li>' +
+            '   <li>' +
+            '       <button class="icon-search zoomin">' +
```
tacruc, line 290: shall we use some more unique class or an id? zoomin might occure more often on the page or?
@eneiluj what do you think?

tacruc:
looks good to me and is working.
@jancborchardt could you check the layout?

src/photosController.js
wronny: as suggested I changed button class to be more unique

tacruc:
Cool, Thanks

PR #333, opened 2020-03-12
photo group in viewer better sorted by date

wronny:
while discussing pros vs cons of #293 it came up that click on a group of photos will show photos in viewer in random order. This is a little fix to show photos instead sorted by date.


wronny:
@tacruc as you are already involved in #293  maybe you like to review this one?

tacruc:
Nice Job.

PR #353, opened 2020-04-07
photo-cluster-max-zoom-preview-slideshow

wronny:
As single photos already have an on mouseover tooltip image preview I added this also to photo clusters with more than one photo at max zoom level. This preview "slideshow" is only active at max zoom level.
2020-04-08 00_18_41-Maps - Nextcloud

### DIFF
--- a/src/photosController.js
+++ b/src/photosController.js
@@ -229,33 +229,42 @@ PhotosController.prototype = {
                         // this loop was previously visible therefore we continue with last visible image
                         $('#imgdiv' + imgIndexAttr ).show();  
                     }
-                    function toolTipImgLoop(maxI){
-                        setTimeout(function(maxI){
-                            // we use visibility of imgdiv to control to stop the following loop
-                            if($('#imgdiv').is(':visible')){
+                    var randomId = Math.random(); // 
+                    $('#imgdiv').parent().parent().attr('randomId', randomId);
+                    function toolTipImgLoop(maxI, randomId){
+                        // will will only continue the loop if randomId is matching randomId stored in imgdiv 
+                        // to prevent running multiple loops in parallel could be caused by to fast mouseout / mouseover events
+                        setTimeout(function(maxI){  // this function content will be executed after timeout of 3 sec (3000 ms)
+                            var randomIdFromImgdiv = $('#imgdiv').parent().parent().attr('randomId');
+                            if ( randomId == randomIdFromImgdiv ){ 
                                 var i = $('#imgdiv').parent().parent().attr('imgindex');
                                 var j = (parseInt(i)+1);
-        		                    if ( i == maxI ){ // if i reached max image j need to start with 0 again to continue with 1st image again
+        		        if ( i == maxI ){ // if i reached max image j need to start with 0 again to continue with 1st image again
                                     j = 0;
                                 }
+                                // now we will fade out the current img and fade in the next image
                                 $('#imgdiv' + i ).fadeOut('fast', function(){ 
                                     $('#imgdiv' + j ).fadeIn('fast'); 
                                 });  
-		                            if ( i == maxI ){ // and now we also need to switch i back to 0 to contine the loop
+		                if ( i == maxI ){ // and now we also need to switch i back to 0 to contine 
                                     i = 0;
                                 }else{
                                     i++;
                                 }
-                                $('#imgdiv').parent().parent().attr('imgindex', i);
-                                toolTipImgLoop(maxI); 
+                                $('#imgdiv').parent().parent().attr('imgindex', i); 
+                                // after storing current value of i (loop img index) in imgdiv imgindex attribute
+                                // we will call again toolTipImgLoop to continue the loop
+                                toolTipImgLoop(maxI, randomId);
                             }
-                        }, 3000, maxI);
+                        }, 3000, maxI); // timeout and variable of above setTimeout
                     };
-                    setTimeout(toolTipImgLoop(maxI), 500, maxI);
+                    // initial call of toolTipImgLoop 
+                    // we will do the inital call of toolTipImgLoop using setTimeout with timeout 500 ms to ensure that tooltipopen has been completed 
+                    setTimeout(toolTipImgLoop(maxI, randomId), 500, maxI, randomId); 
                 });
                 cluster.on( "tooltipclose", function( event, ui ) {
-                    // hide imgdiv on mouseout to stop the img preview loop (see above)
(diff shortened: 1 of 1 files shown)


src/photosController.js
```diff
@@ -177,6 +178,86 @@ PhotosController.prototype = {
...
+                    }
+                    function toolTipImgLoop(maxI){
+                        setTimeout(function(maxI){
+                            // we use visibility of imgdiv to control to stop the following loop
+                            if($('#imgdiv').is(':visible')){
+                                var i = $('#imgdiv').parent().parent().attr('imgindex');
+                                var j = (parseInt(i)+1);
+        		                    if ( i == maxI ){ // if i reached max image j need to start with 0 again to continue with 1st image again
+                                    j = 0;
+                                }
```
tacruc, line 240:
```suggestion
                                var j = (parseInt(i) + 1) % maxI;  // if i reached max image j need to start with 0 again to continue with 1st image again
```

src/photosController.js
```diff
@@ -177,6 +178,86 @@ PhotosController.prototype = {
...
+                                $('#imgdiv' + i ).fadeOut('fast', function(){ 
+                                    $('#imgdiv' + j ).fadeIn('fast'); 
+                                });  
+		                            if ( i == maxI ){ // and now we also need to switch i back to 0 to contine the loop
+                                    i = 0;
+                                }else{
+                                    i++;
+                                }
```
tacruc, line 248:
```suggestion
                               i = (i + 1) % maxI;  // and now we also need to switch i back to 0 to continue the loop
```

src/photosController.js
```diff
@@ -177,6 +178,86 @@ PhotosController.prototype = {
...
+                    if ( typeof imgIndexAttr == typeof undefined || imgIndexAttr == false ) {
+                        // imgindex not yet defined therefore this is the fist time tooltipopen running 
+                        // for this cluster. Therefore we start with imgindex 0
+                        $('#imgdiv').parent().parent().attr('imgindex', '0' ); // saving 0 as start value to attribute imgindex
+                        $('#imgdiv0').show(); // showing first image
+                    }else{
+                        // this loop was previously visible therefore we continue with last visible image
+                        $('#imgdiv' + imgIndexAttr ).show();  
+                    }
+                    function toolTipImgLoop(maxI){
```
tacruc, line 232: Sorry I can not follow your construction here.
What does this function do?
So apparently it sets two timeouts? One is calling the function itself every 500ms?
The other is done after 3 seconds and does the actual action of the function or?
Can you clarify it with some commends, such that a stupid guy like me can understand how the program is working?

src/photosController.js
```diff
@@ -177,6 +178,86 @@ PhotosController.prototype = {
...
+                                    i = 0;
+                                }else{
+                                    i++;
+                                }
+                                $('#imgdiv').parent().parent().attr('imgindex', i);
+                                toolTipImgLoop(maxI); 
+                            }
+                        }, 3000, maxI);
+                    };
+                    setTimeout(toolTipImgLoop(maxI), 500, maxI);
```
tacruc, line 254: I don't get this line setTimeout is requering a function as it first argument, but toolTipImgLoop is not returning a function.
did you mean
'setTimeout(toolTipImgLoop, 500, maxI)'?
I see that this code is working, but I don't get why. Addtionally I'm seeing wired affects, that the loop over the images is not switching with constant intervalls, but sometimes an image is shown 3 seconds and sometimes they switching every half second.

src/photosController.js
tacruc: Just a minor suggestion, you can use modulo "%" for this cyclic counting.

src/photosController.js
tacruc: Just a minor suggestion, you can use modulo "%" for this cyclic counting.

wronny:
Hello tacruc, many thanks for reviewing this PR that detailed. I will try to provide some answers and more details as replies to your comments.
I also add an commit to fix the artefact that I was able to identify based on your feedback - hopefully it is the same artefact ;-)

src/photosController.js
wronny: Thanks for that hint. I tried to change the code to use this kind of modulo for this variable cycling but I was not by doing as suggested because this did cause that the last image was not shown any more. Maybe I will try to investigate on this later again...

src/photosController.js
wronny: I also tried as suggested here but this caused some double preview artefact at the end of the loop. Maybe I will try to investigate on this more detailed later ...

src/photosController.js
wronny: ok, first of all we need to distinguish the two different used setTimeout calls:
- the later one does the initial call of toolTipImgLoop with a small delay of 500 ms just to assure that the tooltipopen has been completed before the loop image loop will start
- the use of setTimeout within function toolTipImgLoop is used to delay all actions inside this function by 3000 ms. I was trying a bit to get the looping delayed using setTimout working but finally I was inspired by https://stackoverflow.com/a/22154172

wronny:
As I mentioned above with the commit 02333176ae056fa692b563f669f9bb8929a5bb6e I hopefully fixed the artefact of faster image loop in case of to fast changing between mouseout and mouseover events. This caused in the before version that two (or maybe more) threads of toolTipImgLoop (started via setTimeout) were fading to the next image in parallel. In the before version I tried to prevent this by stopping the toolTipImgLoop at mouseout (tooltipclose) by hiding the parent imgdiv but based on tacruc's hint I identified via some debugging that in case of to fast changing between mouseout and mouseover again the last setTimeout loop has not been finished before imgdiv got visible again due to new mouseover event. 
Therefore I changed the way how to prevent from having multiple instances of toolTipImgLoop by adding a randomId to every mouseover tooltipopen and keep this randomId in the parent imgdiv as an attribute. Inside the toolTipImgLoop I now verify that the randomId oft the toolTipImgLoop does still match to the randomId kept in imgdiv attr. In case it does not match this instance of the toolTipImgLoop will not continue to loop.

src/photosController.js
tacruc: Wait, it is a maxI+1 if the last image has i=maxI

src/photosController.js
tacruc: Ah now I see, your comments helped thanks. Additionally I misread where the function body was ending. I somehow thought that the second settimeout is in the toolTipImgLoop, which completed my confusion.

tacruc:
Nice :+1:

Issue #346, opened 2020-03-27
Remove photos from map


call-me-matt:
To try out the app, I placed photos randomly on the map. Can't figure out how to remove them from there again (tried renaming the photo directory and reinstalling the app).
Also, I think at the first startup I selected a photo directory but could not find a setting to verify or change.

tacruc:
I guess the photos will come back. You have to remove the geo information form the picture.
With some exif editor.

call-me-matt:
That was what I searched for in the beginning, but could not find any EXIF/location data added to the pictures.

How to reproduce:
* right click on map
* place pictures here
* select folder
* (re)move pictures on map by ...?

wronny:
you can remove a single photo as follows: 
1. zoom to a zoom level that you have a single photo on the map (and not a batch of photos with a count number at the upper right corner of the batch of photos)
2. use right-click menu and choose "remove geo data".
![2020-03-29 16_02_48-Maps - Nextcloud](https://user-images.githubusercontent.com/3709024/77851079-1b691400-71d7-11ea-9de1-8fb887ec2b6e.jpg)
Please note that this right-click menu is currently only available if you're able to reach single photos via zoom level but not if you still have a batch of photos at maximum zoom level.

call-me-matt:
ah, thank you for clarification. I will change the title of this issue, then

wronny:
I also stumbled about this. I'm already thinking about implement this. What would you personally prefer? 
a) to remove all photos of that batch (all the photos you added at once using add photos using a folder)? Or 
b) only to remove single photo(s) of that batch (you added at once using add photos using a folder)?
c) to have a right-click menu that will open a menu tree per photo to allow to "move" or "remove geo data" per photo?

call-me-matt:
Originally, I expected an entry below photos to appear, just like contact groups or tracks. There, I would have thought to remove the added folder again.
If this is too complicated, I would opt for a), as folders might be big and removing each picture individually may be painful.

wronny:
I added a PR to add "Move all" as well as "Remove all geo data" in context menu of photo cluster (more than one photo) at max zoom level :
![2020-04-20 23_06_43-Maps - Nextcloud](https://user-images.githubusercontent.com/3709024/79800058-d6f71100-835b-11ea-92c1-5a650d231d14.jpg)
@tacruc maybe you like to review?

PR #365, opened 2020-04-20
added move and remove for photo cluster at max zoom level

wronny:
as discussed in https://github.com/nextcloud/maps/issues/346 I added "Move all" and "Remove all geo data" to cluster context menu at max zoom level.

### DIFF
--- a/src/photosController.js
+++ b/src/photosController.js
@@ -77,7 +77,7 @@ PhotosController.prototype = {
             that.enterMoveAllPhotoMode();
             that.map.closePopup();
         });
-        $('body').on('click', '.resetphoto', function(e) {
+        $('body').on('click', '.resetphoto', function (e) {
             var ul = $(this).parent().parent();
             var filePath = ul.attr('filepath');
             that.resetPhotoCoords([filePath]);
@@ -389,7 +389,7 @@ PhotosController.prototype = {
         a.layer.openPopup();
     },
 
-    getPhotoClusterContextmenuPopupContent: function(layerId, availZoomLevels) {
+    getPhotoClusterContextmenuPopupContent: function (layerId, availZoomLevels) {
         var viewText = t('maps', 'Show batch in viewer');
         var zoomText = t('maps', 'Zoom in');
         var moveText = t('maps', 'Move all');
@@ -516,7 +516,7 @@ PhotosController.prototype = {
         this.photosController.placePhotos(photoFilePathList, [lat], [lng], false);
     },
 
-    updateTimeFilterRange: function() {
+    updateTimeFilterRange: function () {
         this.updateMyFirstLastDates();
         this.timeFilterController.updateSliderRangeFromController();
     },
--- a/src/photosController.js
+++ b/src/photosController.js
@@ -17,6 +17,7 @@ function PhotosController (optionsController, timeFilterController) {
     this.timeFilterEnd = Date.now();
 
     this.movingPhotoPath = null;
+    this.movingPhotoClusterLayerId = null;
     this.isPhotosInstalled = OCP.InitialState.loadState('maps', 'photos');
 }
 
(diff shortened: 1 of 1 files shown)


tacruc:
The code looks good to me, but I haven't tested. I'm not sure if I will find time before the weekend.

wronny:
> I'm not sure if I will find time before the weekend.

No worries, no hurry

tacruc:
Ok tested it, works nicely. Thanks.
There is a minor thought, I'm not shure how to solve it yet.
The left click button has two behaviors:

- more than x pictures zoom, 
- less than x pictures slideshow

The reight click botton has two different menus:
- not max zoom level
- max zoomlevel

I think we should synchronize them:

1. make both depending on the number of pictures
2. make both depinding on the zoom level 
3. make the right click menu allways the same and keep the left click action as it is.

Any thoughts on this @eneiluj @jancborchardt @wronny ?

wronny:
thought from my point of view are:

- currently "move" as well as "remove geo data" was only available for single photos therefore I added "move all" and "remove all geo data" for the minimum photo cluster count as possible which I was only able to identify by zoom level (at max zoom level). However this guarantee that this can only being used for photos with the same geo data.

- due to the fact that other (outer) zoom level will combine photos from (many) different locations it don't make much sense to allow moving them all to one geo location. And in case of using it by mistake "move all" or "remove all geo data" would be a nightmare if using it form a lower zoom level if hundreds or more photos might be affected 

- I'm not sure if left and right click needs to be synchronized because from my point of view left click is some kind of fast and default behaviour and right click provide some kind of special or detailed options.

tacruc:
Well, I know the details of how things are programmed and I manged to get confused by the current situation. I don't want to know, if I would be able to explain this behavior to my Grandparents. 

And at least a temporally redo action would be nice anyway, or?

wronny:
after a long time being offline I would like to come back to this merge request. The intend of this mr was to fill the following gap: 
In case of having more than one photo at the same location (either because of same geo data or after placing photos on the map using a folder) there is no option available to move those batches of photos or to remove geo data for those batches of photos because the only available right-click menu items for batches of photos are "view" and "zoom in" - even at max zoom level. This has been discussed/complained in #346 as well as #416 . Therefore I added the new right-click menu items "move all" and "remove all geo data" to batches of photos at max zoom level only:
grafik
I limited this to max zoom level to have the same the behaviour as for other photos where "move" and "delete geo data" is only available for single photos. By limiting these menu items to max zoom level these menu items will be limited to the minimum size of photo batches. This also prevents from any usage by mistake at lower zoom levels.
@tacruc , @eneiluj maybe you can review and share your opinion?
thanks

aleksandrmetik:
Hello @eneiluj 
Could you please help us with the review? This feature is really important for us.
Many thanks!

src/photosController.js
```diff
@@ -358,7 +379,8 @@ PhotosController.prototype = {
     photoClusterContextmenu: function(a) {
         var layerId = a.layer._leaflet_id;
         a.layer.unbindPopup();
-        var popupContent = this._map.photosController.getPhotoClusterContextmenuPopupContent(layerId);
+        var availZoomLevels =  this._map.getMaxZoom() - this._map.getZoom();
```
tacruc, line 382: `var groupEditEnabled = a.layer.getChildCount() <= 20 || this._map.getZoom() === this._map.getMaxZoom()`

src/photosController.js
```diff
@@ -367,21 +389,39 @@ PhotosController.prototype = {
         a.layer.openPopup();
     },
 
-    getPhotoClusterContextmenuPopupContent: function(layerId) {
+    getPhotoClusterContextmenuPopupContent: function (layerId, availZoomLevels) {
```
tacruc, line 392: `getPhotoClusterContextmenuPopupContent: function (layerId, availZoomLevels,  groupEditEnabled) {`

src/photosController.js
```diff
@@ -367,21 +389,39 @@ PhotosController.prototype = {
...
-            '   </li>' +
+            '   </li>';
+        if (availZoomLevels > 0){
+            res +=
             '   <li>' +
             '       <button class="icon-search zoomin-photo-cluster-right-click">' +
             '           <span>' + zoomText + '</span>' +
             '       </button>' +
+            '   </li>';
+        }else{
```
tacruc, line 411: `if (groupEditEnabled) {`

tacruc (commented):
This is how I would synchronize the behaviour.

Still keep in mind #510, which might be adopted, too.

kds69:
Can someone please review, this is very annoying missing feature and fix would be great.

tacruc:
Please, don't as it would there is #510 nearly finished.

PR #382, opened 2020-05-16
Fixes to sharing link page

jancborchardt:
- Fix popup not showing anything in favorite shares
- Remove outdated CSS overwriting core styles which have been fixed
- Favorite share: Place the footer in the app-navigation so it is not below the map

What I’d still like to fix, or if anyone can answer the questions:
- [ ] It would look much better if the map would zoom to the bounds of the favorites layer. How is that possible? (I found the `this.map.fitBounds(catLayer.getBounds(), {padding: [30, 30]});` function but don’t know how to get the layer, and neither how to get the map)
- [ ] When logged in, it uses Mapbox layers as configured. But in the share view the standard OpenStreetMap layer is used. Why is that?

### DIFF
--- a/src/PublicFavoriteShare.vue
+++ b/src/PublicFavoriteShare.vue
@@ -86,13 +86,6 @@ export default {
 </script>
 
 <style lang="scss">
-/* Override header style */
-#header {
-    .header-shared-by {
-        color: var(--color-primary-text);
-    }
-}
-
 #content {
     height: 100%;
 }
--- a/src/PublicFavoriteShare.vue
+++ b/src/PublicFavoriteShare.vue
@@ -72,6 +72,7 @@ export default {
 
 	mounted() {
 		this.getFavorites()
+		this.moveFooter()
 	},
 
 	methods: {
@@ -81,6 +82,11 @@ export default {
 			updateFavorite: `${PUBLIC_FAVORITES_NAMESPACE}/updateFavorite`,
 			deleteFavorite: `${PUBLIC_FAVORITES_NAMESPACE}/deleteFavorite`,
 		}),
+		// Place the footer in the app-navigation so it is not below the map
+		moveFooter() {
+			const footer = document.getElementsByTagName('footer')[0]
+			document.getElementById('app-navigation').appendChild(footer)
+		},
 	},
 }
 </script>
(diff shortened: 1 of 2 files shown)


src/PublicFavoriteShare.vue
```diff
@@ -81,23 +82,38 @@ export default {
...
 </script>
 
 <style lang="scss">
-/* Override header style */
-#header {
-    .header-shared-by {
-        color: var(--color-primary-text);
-    }
-}
-
```
paulschwoerer, line 95: Removing this caused the subtitle in the header to be black on NC 17.0.1, don't know if that matters

paulschwoerer:
> It would look much better if the map would zoom to the bounds of the favorites layer. How is that possible? (I found the this.map.fitBounds(catLayer.getBounds(), {padding: [30, 30]}); function but don’t know how to get the layer, and neither how to get the map)

This doesn't seem as trivial as I first thought, as the markers are residing on different layers, based on their category. I guess it would mean to get all layer bounds and add them up somehow

src/PublicFavoriteShare.vue
jancborchardt: Yep, I kept that in mind but it is something which should be fixed in Nextcloud core. And since Nextcloud 19 will be released next week, it’s fine if we don’t have any workarounds for old releases. :) What do you think?

jancborchardt:
> the markers are residing on different layers, based on their category

Hmm, but if I share one set of favorites, doesn’t that mean they are the same category and thus the same layer?

jancborchardt:
Also thanks for the review @paulschwoerer! Just some small fixes on top of your great work :) We could also merge this for now and do further improvements in follow-ups.

jancborchardt:
Merged as it’s important fixes, and opened 2 follow-up pull requests :)
- Favorite sharing: Uses OpenStreetMap layer instead of Mapbox if configured #388
- Favorite sharing: Zoom to bounds of favorites instead of showing whole world map #389

src/PublicFavoriteShare.vue
paulschwoerer: Yeah, that's fine by me. Didn't update my Dev setup in a while :)

PR #386, opened 2020-05-20
My maps

tacruc:
The Idea of the PR is to allow for multiple custom maps, with different content. All information belonging to this map is stored in a corresponding folder. Therefor the map can be shared by technically just sharing the folder.

ToDo List
- [x] MyMapsController
- [x] MyMapsService
- [x] read and wirte Favorits into folder_path/favorits.json
- [x] load pictures only from the folder
- [x] save the options values into folder_path/.maps
- [x] My Maps Menu
  - [x] generate one entry per map
  - [x] open the new map on clicking the entry
  - [x] restoring MyMaps enable status on pageload
  - [x] add new MyMap
  - [x] delete MyMap
  - [x] rename MyMap
  - [x] change MyMap color
- [x] Create an issue to add JSON_THROW_ON_ERROR, when php 7.2 support is dropped


wronny:
I like this approach very much. Inspired by paulschwoerer's fav sharing I was already thinking about sharing tracks. But I like your approach because if e.g. thinking of hiking or cycling etc. I would like to share places, tracks and photos in one combined shared map which will be implemented as far as I understood your approach.
One issues I identified so far in my NC18.0.4.2:

- adding favs ("Your favorites +" as welll as right click "add a favorite") not working with errors
[nextcloud.log](https://github.com/nextcloud/maps/files/4691701/nextcloud.log)

I personally perfer to keep nc in one window instead of opening a new window/tab per click on a your maps entry
Please continue this very good approach!

tacruc:
I guess it is all correlated to using JSON_THROW_ON_ERROR which is introduced in PHP 7.3. What is the minimum PHP version for NC19?

tacruc:
@wronny could you check if it is working now on php7.2.

wronny:
unfortunately still the same error with PHP 7.2.26. I will also test with higher php versions but not before end of next week...

tacruc:
Then I guess I have to install PHP 7.2 somewhere.

tacruc:
@eneiluj what do you think has still to be done?
I'm unshure about error handling, like stoarge not avalibe, handling file not writable errors.
Do you have expierence in this?

eneiluj:
@tacruc Conflicts with current master (with service injection changes) have been solved. I merged master in this branch so it can easily be merged in master when we're done :grin:.

I fixed and changed a few things here and there:
* I think userId should not be given to Services, only to Controllers which pass it to Services if necessary
* There were problems with photo placement
* No photo folder placement in my-maps for the moment (disabled in UI and backend)
* Reduce the number of involuntary option savings on page load (sometimes causing a crash in the backend because `.maps` file is locked)
* Adding a favorite was broken in normal page
* Favorite category state restoring was broken

About your questions...I don't have much experience in error handling but I think you're right. Backend should be more robust. I might look around and do it next time I have time. Maybe we could do it the safest way possible, like checking if `.maps` is writeable on page load and continue only if it is. What do you think?

The most important thing IMO is to properly test and polish all that before merging. I'm pretty sure we left some bugs and undone business.

tacruc:
escapeHTML is undefined in NC20. Do you know a good replacement?

tacruc:
grafik

tacruc:
Follow up PR #556

jancborchardt:
@tacruc nice! I deleted the outdated branch here to clean up, that’s ok right? If not you can always restore it. :)

tacruc:
Actually I would like to keep it, as a backup as long the other is not merged. The Php functions are in the vue rewrite branch. But I'm not shure if the javascrip stuff is integrated.

PR #390, opened 2020-05-22
Replaced array_key_exists with isset

tacruc:
With php 7.4 array_key_exists is deprecated and spams the log.


tacruc:
I found this Regex: `array_key_exists\((.*), (\$[^),]*)\)` usefull
and `isset($2[$1])` for the replacement.

PR #391, opened 2020-05-22
Favorite sharing: Zoom to bounds

wronny:
This is my approach to zoom to bounds in a favorite share by getting min and max latlng over all favorites and fit to these result bounds. Not sure if this is the best approach but maybe a starting point.

### DIFF
--- a/src/components/MapContainer.vue
+++ b/src/components/MapContainer.vue
@@ -59,7 +59,7 @@
 					:icon="createNewDivIcon(categoryKey)"
 					@popupopen="storeCurrentlyOpenPopup(favorite.id)"
 					@popupclose="forgetCurrentlyOpenPopup(favorite.id)"
-					@ready="marker => storeMarkerReference(favorite.id, marker)">
+					@ready="marker => storeMarkerReference(favorite.id, marker, favoriteCategories[categoryKey].length)">
 					<LPopup>
 						<FavoritePopup
 							:favorite="favorite"
@@ -208,6 +208,13 @@ export default {
 		this.featureGroup = null
 		this.mapClickPopupLocked = false
 		this.markerMap = []
+		this.markerCounter = 0
+		// dummy values to replace with min and max values taking from favorites
+		// see storeMarkerReference below
+		this.minLat = 90
+		this.maxLat = -90
+		this.minLng = 720
+		this.maxLng = -720
 	},
 
 	methods: {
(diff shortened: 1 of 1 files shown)


jancborchardt:
@wronny btw, since you are a member of the Nextcloud organization on Github, you can also simply work on branches of this repository and you can get rid of your own fork. :) This makes collaboration on the same branch much easier.

tacruc:
@jan can you try if it works with markers in Japan?  In the Pacific, where the longitude is jumping. I assume that you can set two markers with 1km distance and show the hole world. If the one is at smallest latitude + epsilon and the other is at max latiude - epsilon.

paulschwoerer:
Thanks for your PR @wronny 

While this is definitely a valid approach, I think it kind of reinvents the wheel which might lead to issues as @tacruc pointed out. LeafletJS provides a getBounds() function on nearly all map objects, including layers. Something like this could get the bounds for all favorite markers using onboard LeafletJS functionality.

```
// component MapContainer
getMarkerBounds() {
    const markers = []
    for (const categoryKey in this.favoriteCategories) {
        markers.push(...this.favoriteCategories[categoryKey].map(favorite => marker({
            lat: favorite.lat,
            lng: favorite.lng,
        })))
    }

    // eslint-disable-next-line new-cap
    return (new featureGroup(markers)).getBounds()
},
```

This could then be used with Arne's approach (`map.fitBounds(this.getMarkerBounds(), [30, 30])`).

You could use a similar approach in your PR. Just add the markers to a featureGroup in the `storeMarkerReference()` function and get the bounds of that.

*Please keep in mind, this is all untested*

Let me know what you think

tacruc:
Ok Actually, I think the problem I described seams to  be more fundamental. Usally lng is defined between -180 and +180, but with leaflet it is easy to add favorites add -190 or -270 or the same above. So in our hole maps code, russia scrolling to the right is different point than russia scrolling to the left. 
Therefore we would break stuff if we would compute the right mapbounds here.
Let me give an example
consider a favorite at -170 and one at 170. The minimal "correct" bound to see both favorites would be 170 to -170, but as leaflet it self does not identify the periodicity this would probably result in an error. Therefore we have to show the map from 170 to 190 (which is somehow out of the definition of the lng) and would only show the favorite at 170 as the other one is only drawn once at -170 and leaflet does not identify "190" to be the same as "-170".
Therefore I'm fine to for now to merge the PR as it is. 

I thing we should keep in mind for the future that the world is periodic and we therefore have to draw favorits multiple times. I would also recommend that we should save proper lng values which are between [-180,180] in the future.

jancborchardt:
Ok – so we merge the pull request as is since it fixes the issue. And then @tacruc could you open an issue with that description so it doesn’t get lost? :)

And yep, +1 for showing the favorites multiple times, if you live around the antemeridian (or just have favs both in Asia and the Americas).

wronny:
@paulschwoerer many thanks for your advise and I agree that using featureGroup and getBounds() would be the better approach. However I was not able to successfully add the use of featureGroup.

wronny:
@tacruc yep, looks more and more complicated. When adding favs using NC maps only I have no issue to zoom to bounds (neither with my user nor using this PRs shared link version). However if using favs imported from e.g. OSMAND app I also run into this issue you described (for both via user as well as via shared link). Therefore this possible lng jumping issue is related to the way how cords are saved in the db. Maybe we need to take into consideration the "wrap" related settings and methods of leaflet.

PR #418, opened 2020-07-09
Added sidebar to the viewer

tacruc:

### DIFF
--- a/src/contactsController.js
+++ b/src/contactsController.js
@@ -1,6 +1,7 @@
 import { generateUrl } from '@nextcloud/router';
 
 import { basename, formatAddress } from './utils';
+import escapeHTML from 'escape-html';
 
 function ContactsController (optionsController, searchController) {
     this.contact_MARKER_VIEW_SIZE = 40;
--- a/src/photosController.js
+++ b/src/photosController.js
@@ -2,6 +2,8 @@ import { generateUrl } from '@nextcloud/router';
 
 import { basename } from './utils';
 
+import escapeHTML from 'escape-html';
+
 function PhotosController (optionsController, timeFilterController) {
     this.PHOTO_MARKER_VIEW_SIZE = 40;
     this.photosDataLoaded = false;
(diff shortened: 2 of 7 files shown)


templates/index.php
```diff
@@ -21,6 +21,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>
  *
  */
+script('maps', 'mockupFilesApp');
```
eneiluj, line 24: What is this? It still works if I remove it.

src/contactsController.js
```diff
@@ -1,6 +1,7 @@
 import { generateUrl } from '@nextcloud/router';
 
 import { basename, formatAddress } from './utils';
+import escapeHTML from 'escape-html';
```
eneiluj, line 4: Could you remove those which are out of topic?

src/photosController.js
```diff
@@ -205,7 +207,7 @@ PhotosController.prototype = {
                 iconUrl = _app.getImageIconUrl();
             }
             var label = cluster.getChildCount();
-            if( availZoomLevels == 0 && label > 1){ 
+            if( availZoomLevels == 0 && label > 1){
```
eneiluj, line 210: Could you also remove those changes on trailing white spaces? Or maybe fix the code style in this file?

This line could be:
``` javascript
if (availZoomLevels === 0 && label > 1) {
```

templates/index.php
tacruc: I thougt I allready removed it.

src/contactsController.js
tacruc: My problem is that without this my dev setup is not working, as it is running NC20. But I can bring it up in a own PR.

tacruc:
> Uncaught TypeError: can't access property "extend", OCA.Files.DetailTabView is undefined
> Uncaught TypeError: can't access property "Client", t.Files is undefined
> Uncaught TypeError: can't access property "extend", OCA.Files.DetailTabView is undefined

I thing the errors are a problem we are not going to solve now. They happen for the photos app, too. I think it is related to different tabs which are not yet migrated.

eneiluj:
Ok so ready to merge after having removed the `mockupFilesApp` script. Tell me if you don't have time to do it, I'll do it.

skjnldsv:
> I thing the errors are a problem we are not going to solve now. They happen for the photos app, too. I think it is related to different tabs which are not yet migrated.

Yep, you can ignore those :)

PR #420, opened 2020-07-10
Added Regex to detect coordinates in the search query.

tacruc:
Unfortunattly I do net get the results merged with the results from Nominatim.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -1764,7 +1764,7 @@ import { brify, getUrlParameter, formatAddress } from './utils';
             this.search(str, this.handleSearchResult, this);
         },
 
-        handleSearchResult: function(results, that) {
+        handleSearchResult: function(results, that, isCoordinateSearch = false, searchString = '') {
             if (results.length === 0) {
                 OC.Notification.showTemporary(t('maps', 'No search result'));
                 return;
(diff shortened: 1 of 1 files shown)


tacruc:
@eneiluj could you have a look?
I managed to get it working if the length of results is 1.
If the length of results is longer then one I followed the program until line 1789 in script.js but I the it finishes without showing something.

eneiluj:
@tacruc Yep, could you provide 2 examples, one with a single result and one with multiple?

tacruc:
geo:0.0,0.0 single result
0.0,0.0  two results

eneiluj:
Ok, got it, the autocomplete list does not show up because no label/value matches the search string in the result list.

I added the search string in the "fake" coordinate result label and in the nominatim results when we know it's a coordinate search. "0.0,0.0" now displays both the "Point" and "Soul Buoy".

eneiluj:
@tacruc I added a last thing: when choosing a "coordinate" result from multiple ones, no result popup was displayed.

If it's ok for you we can merge that and make the release.

tacruc:
You have to review, I cant :D

eneiluj (approved):

PR #453, opened 2020-09-25
fix declaration of object_uri DB field

eneiluj:
This was making the installation crash on instance where Maps was never installed (all migration scripts running).

Strange that we didn't bump into that earlier. Anyway, old migration scripts run fine on fresh install now.

`object_uri` column is always affected when inserting/updating `maps_address_geo` table anyway.


tacruc (approved):

aszlig:
@eneiluj: Do you intend to backport this to the Nextcloud 19 release as well? Version 19.0.4 has been released and installing `maps` now results in the same issue.

PR #461, opened 2020-09-29
Temporary fix for navigation style

eneiluj:
This is bringing the "classic" colors back.

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -3,6 +3,20 @@
     z-index: 100000;
 }
 
+#app-navigation > ul > li.active,
+#app-navigation > ul > li a.active,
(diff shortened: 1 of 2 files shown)


tacruc:
For me it is still not working for the contact categories.

eneiluj:
@tacruc Thanks for the tests. Should be fine now.

tacruc:
Your removed the blue color completly or?
Fine with me.

tacruc (approved):

eneiluj:
Well I did the simplest fix i could because i didn't want to struggle. It now looks like in NC 19. We'll probably use the default "active" color with Vue navigation components so it will use the theme color.

PR #465, opened 2020-10-01
Address Service: Convert Open Location Codes to lat/lon

earboxer:
Nominatim folks suggest this be implemented by client sites, rather than on their end (https://github.com/osm-search/Nominatim/issues/914)


tacruc:
Looks good to me, I will test it soon.

@nextcloud/maps   What do you think? Do we have problems with additional dependencies?

tacruc:
@earboxer: Testing now I noticed, that you forgot to commit the composer.lock file.

tacruc (approved):

PR #510, opened 2020-12-07
[WIP] Rewrite UI in Vue.js

eneiluj:
Hey there!

Why not fully rewriting Maps UI :grin:?

Sorry in advance for the size of this text. It's also a memo.

### How to try it
There isn't any DB migration steps involved. It's safe to switch to this branch and then get back to master. Just run `make dev` to build.

### Immediate goals:
* get rid of JQuery
* make the interface more stable/solid
* make it easier to understand/improve the interface
* simplify tile layers management => implement custom tile layers management in settings

### Long term goals:
* reuse some components in other apps like Photos
* produce independent components able to access Maps data from outside Nextcloud

### Downsides:
While this banch is incompatible with some of the pending PRs, I think it won't be too hard to adapt them and benefit from using Vue.

### Todo
* [x] base map
* [x] routing
* search
    * [x] for user data
    * [x] for places
    * [x] for points of interest
    * [x] on left click
* [x] track me
* [x] locate control
* [x] filter slider
* [x] photos
* [x] contacts
* [x] favorites
* [ ] tracks
* [ ] devices
* [ ] my maps

### Map library
Vue2Leaflet has been chosen over mapbox-gl-vue and vuelayers (OpenLayers) for various reasons.
* It's quite hard to make Mapbox vector tiles work in OpenLayers
* Avoid loosing Leaflet routing machine which only works with Leaflet and is compatible with many routing engines.
* Avoid loosing all the other Leaflet plugins we use (easybutton, markercluster, elevation chart, context menu...)

### Overall structure
The code structure is quite different than before. It used to be grouped by topic (photos, favorites...) and each controller took care of everything (display stuff on the map, manipulate the data, make network requests etc...). Communication between controllers was mostly done by method calls.
All the data is now managed by the `App` top component and then passed down to navigation menu and Map stuff. UI Entities are now more independent from each other. Almost all communication is now done via events.

### Action history
Photo and favorite actions are now stored and it's possible to cancel and redo them. There's a new map control (top right) with cancel and redo buttons.

### SearchField
The new SearchField component is used in the search bar and in each routing step. Routing steps used to be "augmented" with JQuery autocompletion. `Multiselect` (from nextcloud-vue) is now used everywhere. As it was not possible to inject a Vue component in the leaflet-routing-machine control, things have changed there. Integrated step management is now hidden and replaced by a RoutingSteps component which works in parallel of the routing machine. The routing machine itself has been abstracted in a RoutingMachine component.

### Moving favorites and photos
Before: Context menu -> click move -> the move cursor changes to a crosshair -> click on target location to move the favorite/photo
After: Enable dragging in navigation menu -> drag'n'drop favorites/photos


src/components/AppNavigationFavoritesItem.vue
```diff
@@ -0,0 +1,247 @@
...
+				@update:title="$emit('rename-category', { old: catid, new: $event })">
+				<template #icon>
+					<div :class="{ favoriteMarker: true, favoriteMarkerDark: isDarkTheme }"
+						:style="'background-color: #' + c.color" />
+				</template>
+				<template slot="counter">
+					&nbsp;
+					<span v-if="enabled && nbFavorites && c.enabled">{{ c.counter }}</span>
+				</template>
+				<template slot="actions">
```
tacruc, line 67: I think some action to change the color is missing here

src/components/Sidebar.vue
```diff
@@ -0,0 +1,76 @@
...
+		:active="activeTab"
+		@update:active="onActiveChanged"
+		@close="$emit('close')">
+		<!--template slot="primary-actions" /-->
+		<AppSidebarTab
+			id="sharing"
+			icon="icon-shared"
+			:name="t('maps', 'Sharing')"
+			:order="1">
+			PLOP
```
tacruc, line 16: Nice

src/components/Sidebar.vue
```diff
@@ -0,0 +1,76 @@
...
+		</AppSidebarTab>
+		<AppSidebarTab
+			id="settings"
+			icon="icon-settings-dark"
+			:name="t('maps', 'Settings')"
+			:order="2">
+			LALA
```
tacruc, line 23: Awesome

tacruc (commented):
Nicely Done :+1:

src/components/AppNavigationFavoritesItem.vue
eneiluj: It has never been possible to change favorite category color :grin:. The backend does not handle that yet.

src/components/Sidebar.vue
eneiluj: I see that you like the amazing new sidebar features!

src/components/AppNavigationFavoritesItem.vue
tacruc: But then before this the color was random. Now I had 3 categories with the same color.

src/components/AppNavigationFavoritesItem.vue
```diff
@@ -0,0 +1,247 @@
...
+
+::v-deep .icon-save {
+	background-color: var(--color-main-text);
+	mask: url('../../img/save.svg') no-repeat;
+	mask-size: 16px auto;
+	mask-position: center;
+	-webkit-mask: url('../../img/save.svg') no-repeat;
+	-webkit-mask-size: 16px auto;
+	-webkit-mask-position: center;
+}
```
paulschwoerer, line 246: There's a lot of repeated rules here, wouldn't it be nice to only set the values once for all icons?

src/components/map/PhotosLayer.vue
```diff
@@ -0,0 +1,284 @@
...
+		LTooltip,
+		LPopup,
+		ActionButton,
+	},
+
+	props: {
+		map: {
+			type: Object,
+			required: true,
+		},
```
paulschwoerer, line 90: It's been a while since I've worked with Vue, so please correct me if I'm mistaken here. My understanding is, that Vue will set up watchers for all fields of a prop. A change on one of the fields would then trigger a rerender of the component. Do you think this could get quite costly when passing in the whole Leaflet Map instance?

src/components/map/routing/RoutingMachine.vue
```diff
@@ -0,0 +1,526 @@
...
+import 'leaflet-control-geocoder/dist/Control.Geocoder.css'
+import 'leaflet-routing-machine/dist/leaflet-routing-machine'
+import 'leaflet-routing-machine/dist/leaflet-routing-machine.css'
+
+import * as network from '../../../network'
+import optionsController from '../../../optionsController'
+
+export default {
+	name: 'RoutingMachine',
+
```
paulschwoerer, line 22: Is there a reason why you implemented the RoutingMachine as a Vue component, even though it has no associated template?

paulschwoerer (commented):
Good job! That looks like it has been a ton of work. I left a few comments, but couldn't dive deeper, yet. Maybe it would make sense to leverage the vuex store for data loading as has been done for the public favorites?

src/components/AppNavigationFavoritesItem.vue
eneiluj: Yes indeed. I just struggled a bit with style rules range and stopped touching them when it worked :grin:. There sure is a lot of room for factorization.

src/components/map/PhotosLayer.vue
eneiluj: I don't think it re-renders PhotoLayers when a map attribute changes. The mechanism operates at a finer level. It will only re-render what is affected, what uses the changed values.

src/components/map/routing/RoutingMachine.vue
eneiluj: Yes, the reason is it's now easy to plug it in a Leaflet-Vue based map application. Also it can now be treated as a classic component. It has a reactive `visible` prop. Stuff like that.

eneiluj:
@paulschwoerer Thanks for the review.

I never got interested in Vuex. If you convince me I'll probably see how it can be done with Vuex.

> But then before this the color was random. Now I had 3 categories with the same color.

The color generation is supposed to be the same as before. The 2 first category name letters are used to determine the color. Anyway we need to improve this. What stops us to make a move here is that categories are not really stored in the DB, they are just a string field of the favorite table. It would be way better to have a category table and to use its ID as a foreign key in favorites table. I'm going to first implement everything like it was before. That's one of the first things to do after that.

paulschwoerer:
Sorry for not getting back to this, I'll try to do a more in-depth review today or tomorrow, if desired.

paulschwoerer:
So, just had a look at it. A few things:

* It seems that favourite sharing is broken, at least for me. When adding a favourite location and then clicking on "Create share link", I'm getting a 400 error, which seems to originate from [here](https://github.com/nextcloud/maps/blob/8a8fbfc4c557b31ab7005dc7ff6bd56207e576db/lib/Controller/FavoritesController.php#L186). Can you think of a reason, why countFavorites() returns 0 here?

* The Map component is far too large for my personal taste. In think it would be nice to break it up somewhat

* The icon in the top right corner should maybe be a lighter colour or use a background, as it's hard to see, especially on the satellite view.

* I also think it would be nice to have cleaner styling on the place popups.

Other than that LGTM!

eneiluj:
@paulschwoerer Thanks a lot for the review! I agree and acknowledge everything you said :grin:. I'll get back to Maps development once I manage to get less busy...

szaimen:
@eneiluj any update on this? :)

jmporchet:
Please make this PR a reality 😄 
Thanks for your work!!

szaimen:
For anyone that wants to test this branch and doesn't have a developer instance and doesn't want to go through it:
If you have docker installed on your Linux PC, you can simply run the following command:
```
docker run -d \
-e MAPS_BRANCH=enh/vue-rewrite \
-e VIEWER_BRANCH=master \
--name nextcloud-test \
-p 8443:443 \
-e TRUSTED_DOMAIN=192.168.178.30 \
ghcr.io/szaimen/nextcloud-onedev:latest
```
Afterwards the server will be reachable via https://localhost:8443 and you can login with the user `admin` and password `nextcloud`.

---

See further instructions at https://github.com/szaimen/nextcloud-test#nextcloud-easy-test-instance

tacruc:
Ok I made two approaches to improve the interplay between the viewer and the sidebar, therefore there are two branches:

### enh/vue-image-sidebar
The sidebar is provided by the maps app, and the viewer can control it.
Working:
 - [x] Open sidebar from Viewer
 - [x] Viewer does not overlap with sidebar, but opens besides it
 - [x] Viewer expands if sidebar is closed
 - [x] sidebar goes into fullscreen mode, when viewer is open
Not working yet/ ToDo:
 - [ ] Program a tab with information about files
 - [ ] For some reason the viewer can only open the sidebar, if it was opened once before from the maps app, -> Quick hack load the maps app with opened sidebar.

### enh/enh/vue-sidebar.js-only
Here the sidebar is generated by the files app, the viewer and the maps app can controll it.
Working:
 - [x] Interaction between sidebar and viewer
 - [x] Full File information Images and tracks
 - [x] Sharing images and tracks from the maps app
 - [x] Commenting images and tracks form the maps app
 - [x] Every other feature using the files app sidebar

Not Working:
 - [ ] Restoring the options, as we have to create the maps vue app before the sidebar.js gets executed on DOMContentLoaded 
 - [ ] Favorite Sidebar
 - [ ] Track details, but if done probably this would then work in the files app too

tacruc (commented):
The place contact button
image
does not work.

```
[Vue warn]: Error in mounted hook: "Error: iconUrl not set in Icon options (see the docs)."

found in

---> <LMarker>
       <PlaceContactPopup> at src/components/map/PlaceContactPopup.vue
         <LMap>
           <Map> at src/components/Map.vue
             <AppContent>
               <Content>
                 <App> at src/views/App.vue
                   <Root> vue.runtime.esm.js:619

```

eneiluj:
@tacruc I've had a look at all your changes on top of this branch and I'm fine with them :+1: . You can approve this PR whenever you feel like it's ready and let's merge it. Improvements will be much easier then. I can help and review the future my-maps PR :grin:.

eneiluj:
@tacruc In this branch. I'll check the others later.

tacruc:
So I would suggest merging #700 and #701 as they merge into here and then merging this into master.

tacruc (approved):

szaimen:
🎉🎉🎉🎉🎉

PR #521, opened 2020-12-28
CHG: Tweak nominatim results to get exact building

koying:
By default, Nominatim weight the results of addresses lookup so that the street has precedence over a precise building.
This change tweaks it so that if a specific building is available in the results, it is picked instead.


tacruc:
Thanks

koying:
Hehe... Never too late is it 😃

PR #556, opened 2021-03-08
[WIP] Enh/vue rewrite my maps

tacruc:
Just to keep track of the divergence.


jmporchet:
This seems like a worthwile addition!
I'm not well versed in PHP or Vue, so I'm just crossing my fingers that development continues and that it gets shipped to production in a future version!
Thanks for your work, you guys are awesome @tacruc and @eneiluj !

tacruc:
Lets directly merge #731

Issue #541, opened 2021-02-23
The "unique" column option is not supported


J0WI:
Error: Doctrine\DBAL\Schema\Exception\UnknownColumnOption: The "unique" column option is not supported.

<details><summary>details</summary>
```
 0. /var/www/html/3rdparty/doctrine/dbal/src/Schema/Column.php line 84
    Doctrine\DBAL\Schema\Exception\UnknownColumnOption::new("unique")
 1. /var/www/html/3rdparty/doctrine/dbal/src/Schema/Column.php line 68
    Doctrine\DBAL\Schema\Column->setOptions({notnull: true,length: 255,unique: true})
 2. /var/www/html/3rdparty/doctrine/dbal/src/Schema/Table.php line 355
    Doctrine\DBAL\Schema\Column->__construct("adr", Doctrine\DBAL\Types\StringType {}, {notnull: true,length: 255,unique: true})
 3. /var/www/html/custom_apps/maps/lib/Migration/Version000009Date20190625000800.php line 43
    Doctrine\DBAL\Schema\Table->addColumn("adr", "string", {notnull: true,length: 255,unique: true})
 4. /var/www/html/lib/private/DB/MigrationService.php line 455
    OCA\Maps\Migration\Version000009Date20190625000800->changeSchema(OC\Migration\SimpleOutput {}, Closure {}, {tablePrefix: "oc_"})
 5. /var/www/html/lib/private/DB/MigrationService.php line 418
    OC\DB\MigrationService->migrateSchemaOnly("latest")
 6. /var/www/html/lib/private/Installer.php line 163
    OC\DB\MigrationService->migrate("latest", true)
 7. /var/www/html/apps/settings/lib/Controller/AppSettingsController.php line 448
    OC\Installer->installApp("maps")
 8. /var/www/html/lib/private/AppFramework/Http/Dispatcher.php line 218
    OCA\Settings\Controller\AppSettingsController->enableApps(["maps"], [])
 9. /var/www/html/lib/private/AppFramework/Http/Dispatcher.php line 127
    OC\AppFramework\Http\Dispatcher->executeController(OCA\Settings\Con ... {}, "enableApps")
10. /var/www/html/lib/private/AppFramework/App.php line 157
    OC\AppFramework\Http\Dispatcher->dispatch(OCA\Settings\Con ... {}, "enableApps")
11. /var/www/html/lib/private/Route/Router.php line 302
[...]

Issue #567, opened 2021-03-18
Fresh installation of Maps api endpoints returning 500 errors


localjo:
I have a fresh Nextcloud installation (21.0.0), and I've just installed Maps (0.1.8).

I ran into an issue with database columns when installing, and used the workaround steps here: https://github.com/nextcloud/maps/issues/541

The front end UI looks like it's working fine, but all of the backend API endpoints are returning 500 errors. I will continue debugging, but I suspect there is a database issue at the root of the installation and server errors.

![Screen Shot 2021-03-18 at 10 17 57 PM](https://user-images.githubusercontent.com/516388/111699893-fb43c200-8838-11eb-9c9c-3bbe7b31b7bb.png)

localjo:
This was resolved with #568

PR #568, opened 2021-03-19
Fix db-related install problems on NC 21

eneiluj:
I guess upgrading NC dependencies causes the 'unique' problem, this has probably been removed from the DB connector.

But I don't know where the 'contact_uid' problem comes from. Anyway it can be fixed by not adding it as we get rid of it later anyway.

We could publish after that as it apparently happens with all DB types.

refs #541

### DIFF
--- a/lib/Migration/Version000009Date20190625000800.php
+++ b/lib/Migration/Version000009Date20190625000800.php
@@ -42,12 +42,10 @@ class Version000009Date20190625000800 extends SimpleMigrationStep {
 			$table->addColumn('adr', 'string', [
 				'notnull' => true,
 				'length' => 255,
-				'unique' => true,
 			]);
 			$table->addColumn('adr_norm', 'string', [
 				'notnull' => true,
 				'length' => 255,
-				'unique' => true,
 			]);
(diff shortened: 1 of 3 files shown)


tacruc:
Changes are looking good. Lets wait for some feedback whats happening on systems, where the old broken Migration was already executed.

localjo:
This seems to work for the install, but has no effect on #567 for me (but that issue may not be related).

eneiluj:
@localjo Thanks for the tests and feedback. 

Could you check if `oc_maps_*` tables exist?
If they don't, could you disable the maps app, then delete every row with `appid = "maps"` in oc_migration, then enable the app again? This should create the tables...

I think the 500 errors happen because the tables were not created (because of what's fixed in this PR) and then Nextcloud does not execute the DB migration steps again because they were already done (obviously :grin:).

tacruc (approved):

eneiluj:
@localjo Thanks a lot for the feedback.

VoOoUoOoV:
GJ 
Seems to work! `oc_maps_*` missed in the DB. 
THX!

gmeks:
Works for me to.

Issue #574, opened 2021-03-24
Nc21 tracks keeps loadings sign


Tazzios:
I have 2 Nc setups one with 20 and one with 21. In the 20 the tracks were shown in the 21 the same tracks was not. I now just update my NC20 environment and now the tracks also won`t load there anymore.
For one site i use hanssonit vmware image the other on a regular shared hosting.

I hope i did not overlook an already existing issue.

dasarne:
Same here. 
I use the docker-image (nextcloud 21.0.0) with archlinux. 
./occ maps:scan-photos and ./occ maps:scan-photos and even starting cron.php doesn't change this behavior.

realmicu:
Exactly the same problem. I'm using official docker image with NC 21.0.0 (fpm) and maps 0.1.8.
Looking at browser console, it show "Uncaught SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data". Apparently GPX files no longer get processed correctly.

tacruc:
Can you paste the received tracks data. Somewhere in the browsers network monitor should be a get request for tracks.

Tazzios:
Not sure what you are looking for but this is the only thing with track in the firefox console
`Uncaught SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data
    jQuery 2
    addTrackMap tracksController.js:294
    getTracks tracksController.js:422
    jQuery 7
    getTracks tracksController.js:410
    toggleTracks tracksController.js:153
    restoreOptions script.js:284
    jQuery 7
    restoreOptions script.js:124
    <anonymous> script.js:2139
    jQuery 12
    Webpack 6`

other things maybe relevant
`De bron van ‘https://xxx/apps/files/’ is geblokkeerd vanwege een niet-overeenkomend MIME-type (‘text/html’) (X-Content-Type-Options: nosniff).`

`Content Security Policy: De instellingen van de pagina blokkeerden het laden van een bron op inline (‘script-src’).`

`Handlebars is deprecated: please ship your own, this will be removed in Nextcloud 20`

realmicu:
@tacruc:
Using network monitor, here is what I get for 'GET apps/maps/tracks' request - response payload is:
[...]

PR #587, opened 2021-04-12
Fix loading of tracks

Ablu:
The flow seems to be:
```
occ maps:scan-tracks
```
adds the tracks to the db with `$metadata = ""`. Then, when the UI
initially requests a specific track the metadata is generated. This
however never happens since the UI code attempts to parse the metadata
as JSON before. Fixed by checking for an empty string before (in which
case we return an empty object (the metadata will be read again later
when displaying an individual track)).

I am not sure why the metadata is read here it all. Maybe the parsing
call be be removed completly if it happens again later anyway?

Fixes: #574

### DIFF
--- a/src/tracksController.js
+++ b/src/tracksController.js
@@ -291,7 +291,7 @@ TracksController.prototype = {
             html: ''
         });
         this.tracks[track.id] = track;
-        this.tracks[track.id].metadata = $.parseJSON(track.metadata);
+        this.tracks[track.id].metadata = track.metadata === '' ? {} : $.parseJSON(track.metadata);
         this.tracks[track.id].icon = L.divIcon(L.extend({
             html: '<div class="thumbnail"></div>​',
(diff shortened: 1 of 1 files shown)


Julianoe (approved):

ngt-github:
I stubled upon the same Problem. However, unaware of this pullrequest, I did my own fix using try/catch in order to catch malformed json in general:

replaced line 294: this.tracks[track.id].metadata = $.parseJSON(track.metadata);
with
`		try {
			this.tracks[track.id].metadata = $.parseJSON(track.metadata);
		}catch(e) {
			this.tracks[track.id].metadata = {};
		}`

I am not sure if it is neccessary at all and if there will be cases of malformed json besides an empty string but I just hope for a fixed version soon. This pullrequest works for me too,

eneiluj (approved):
Thanks!
This will be replaced by the Vue.JS rewrite soon anyway.

PR #586, opened 2021-04-11
Fix for division by 0 errors. resolves #645

kovge:
- Fix Division by 0 Errors in case of incorrect (or fake, missing) Geocoordinate data. 
  In general: All division operatios must be handled since php throws Error exceptions.
- getExifPelBackup could throw some type of Error Exceptions, all are Catched as Throwable.
 In general:  php Throws error exceptions as well (previously they were only warnings or notices) They must be handled like exceptions.
- avoid to have images with 0,0 coordinates

PR #648, opened 2021-09-07
prevent error on invalid exif data

### LINKED ISSUE #361: Division by zero error when adding photos
So I'm getting the below error under the following circumstances:
[PHP] Error: Division by zero at /var/www/html/custom_apps/maps/lib/Service/PhotofilesService.php#407

I have a bunch of photos uploaded to NextCloud. I then add them to the scan job with the following command:
docker exec --user www-data nextcloud php occ maps:scan-photos

The next time the cron job runs, I get the division by zero error, and none of the photos appear on the map.
The error is not happening for photos that I've recently uploaded through the Android app, and these seem to be added to the map just fine.
The photos giving the error were taken previously, and uploaded via the Linux desktop app.

Thanks

bananer:
Fixes #361

### DIFF
--- a/lib/Service/PhotofilesService.php
+++ b/lib/Service/PhotofilesService.php
@@ -410,7 +410,9 @@ class PhotofilesService {
                 $pos = strpos($value, '/');
                 if ($pos !== false){
                     $temp = explode('/',$value);
-                    $gps[$key] = $temp[0] / $temp[1];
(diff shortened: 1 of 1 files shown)


njarun (approved):

tacruc:
Can you check if #586 solved your issue?

tacruc:
Assumed to be solved by #586

PR #660, opened 2021-11-16
Webpack: force node sass due to /deep/ modifiers in scss source.

rotdrop:
The style files contain /deep/ in some places which is not understood
dart-sass. OTOH dart-sass is preferred by npm. So rather force webpack
to use node-sass.

Signed-off-by: Claus-Justus Heine <himself@claus-justus-heine.de>


tacruc:

PR #663, opened 2021-11-16
BUGFIX: do not use nominatim "display name" because it is not a displ…

rotdrop:
…ay name.

See upstream issue https://github.com/osm-search/Nominatim/issues/213

Signed-off-by: Claus-Justus Heine <himself@claus-justus-heine.de>


tacruc:

PR #688, opened 2022-02-14
Replace use of IConfig with IMemcache to store the last address lookup

tcitworld:
Using memcache (Redis, memcached, …) should be more efficient than writing to the database. `IMemcache` fallbacks to `ArrayCache` if none is defined.


tacruc:
Wouldn't it make sense then, to not store the time of the last lookup, but set a bool value like AddressLookupBlocked, with a time to live of 1 second. Or is the value not reliable deleted after the ttl?

Between which processes it the Memcache and the fallback ArrayCache shared? It should be shared between users and  session.

tacruc (approved):

PR #617, opened 2021-06-05
Skip a folder if it contains a file named .nomedia or .noimage

schklom:
I haven't tested it, but the code is simple enough. Hopefully it works :)

PR #699, opened 2022-03-16
[Draft]  Skip folders which contain a .nomedia or .noimage file

schklom:
@tacruc Follow-up on https://github.com/nextcloud/maps/pull/617


schklom:
@tacruc DCO is still unhappy, and wants me to do the same commands as in https://github.com/nextcloud/maps/pull/617:

> To add your Signed-off-by line to every commit in this branch:
>
> 1. Ensure you have a local copy of your branch by checking out the pull request locally via command line.
> 2. In your local branch, run: git rebase HEAD~2 --signoff
> 3. Force push your changes to overwrite the branch: git push --force-with-lease origin master

 This should lead (again) to a large volume of commits as in the previous PR. What should I do?

tacruc:
On top it explains how to commit changes, such that the commit is sign. 
> There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.
> 
> To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).
> 
> Here is how to fix the problem so that this code can be merged.

schklom:
Ok i'll try again. But this is really strange because I did the pull request using github.com from the browser, I didn't expect github.com would cause problems.

tacruc:
Actually I think this PR won't do what you would expect. 
It will only exclude images during the inital scan. As soon files within excluded folders are uploaded or changed, they will appear.

PR #700, opened 2022-03-21
Enh/vue image sidebar

tacruc:
This PR improves the interplay of the viewer with the nextcloud Sidebar.

Working and tested:

- [x] Open sidebar from Viewer
- [x] Viewer does not overlap with sidebar, but opens besides it
- [x] Viewer expands if sidebar is closed
- [x] sidebar goes into fullscreen mode, when viewer is open
- [x] Program a tab with information about files

To Solve
- [x] The viewer thinks, that the sidebar is opened on initial page load, even if it's not. Therefore opening the sidebar form the viewer menu only works, if the sidebar was closed before.

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -51,7 +51,7 @@ class PageController extends Controller {
      * @NoCSRFRequired
      */
     public function index(): TemplateResponse {
-        $this->eventDispatcher->dispatch(LoadSidebar::class, new LoadSidebar());
+//        $this->eventDispatcher->dispatch(LoadSidebar::class, new LoadSidebar());
         $this->eventDispatcher->dispatch(LoadViewer::class, new LoadViewer());
 
         $params = array('user' => $this->userId);
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
 				"@nextcloud/auth": "^1.3.0",
 				"@nextcloud/axios": "^1.9.0",
 				"@nextcloud/dialogs": "^3.1.2",
+				"@nextcloud/event-bus": "^2.1.1",
 				"@nextcloud/l10n": "^1.4.1",
 				"@nextcloud/moment": "^1.1.1",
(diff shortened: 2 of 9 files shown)


eneiluj (commented):
Awesome ! Could you make it so the sidebar closes when the image viewer is closed?

tacruc:
> Awesome ! Could you make it so the sidebar closes when the image viewer is closed?

I could try, but then there is no way to open the sidebar for images, on the side of the map.

tacruc:
Ok I remember why I did this.
The "open sidebar" menu element in the top right of the viewer, is only show, if the sidebar was opened once before.
I don't know why. This issue is not noticeable if the sidebar is oped on pageload or together with the viewer. As you can not reach the state, where the viewer is opened, but the sidebar was never opened.

The Issue is resolved if the sidebar is opened and closed.

tacruc:
Ok I think the Issue with the viewer is, that it some thinks that the sidebar is opened on pageload.
I'm not sure, why as `isSidebarShown: false` from https://github.com/nextcloud/viewer/blob/dcf96f25b85b8046da4c5d8e0365106e6970672f/src/views/Viewer.vue#L185 

Anyhow 
`emit('files:sidebar:closed')`
resolves this issue.

We just need to emit it at the correct time.
In the following places it is to early:
- DOMContentLoaded
https://github.com/nextcloud/maps/blob/79e08a6db11ffed5a782ee344ced16a4a3587fcb/src/main.js#L63
- APP.created() 
https://github.com/nextcloud/maps/blob/79e08a6db11ffed5a782ee344ced16a4a3587fcb/src/views/App.vue#L494
- APP.mounted() 
https://github.com/nextcloud/maps/blob/79e08a6db11ffed5a782ee344ced16a4a3587fcb/src/views/App.vue#L508

Any other suggestions, where it might be a good place and time

tacruc:
> Right, i think the same, then the only change to make is to **not** open the sidebar when an image is clicked and the viewer is opened.

It's done :-D

Issue #689, opened 2022-03-02
Maps 0.1.10 messing my debug log with deprecated app.php message


l1gi:
Hi,

I am observing this for some time for number of other NC apps, but this is the last one which is doing this on my instance on Nextcloud 23.0.1. The message follows:

/appinfo/app.php is deprecated, use \OCP\AppFramework\Bootstrap\IBootstrap on the application class instead.

Screenshot attached. Sorry for localized text (Uroven ~ level, Aplikace ~ application, Zprava ~ message, Cas ~ datetime, Ladici ~ debug).

Let me know if you need more info on this.

Regards, -m

![Screenshot from 2022-03-02 09-42-00](https://user-images.githubusercontent.com/17202643/156326627-88d7b3d3-3cb9-4751-92a8-7e2f49c33b5b.png)

PR #703, opened 2022-03-23
Migrated to App Bootstrap

tacruc:
app.php is deprecated (https://docs.nextcloud.com/server/latest/developer_manual/app_development/bootstrap.html)


tacruc:
Fixes #689

eneiluj:
`appinfo/application.php` should be removed. Better to check if everything still works fine without it (file hooks and feature policy).

PR #707, opened 2022-03-24
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/src/contactsController.js
+++ b/src/contactsController.js
@@ -416,8 +416,8 @@ ContactsController.prototype = {
         for (var i = 0; i < contacts.length; i++) {
 
             var geo = [];
-            if (contacts[i].GEO.substr(0,4) === "geo:") {
-                geo = contacts[i].GEO.substr(4).split(",");
+            if (contacts[i].GEO.slice(0,4) === "geo:") {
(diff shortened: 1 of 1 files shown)


tacruc:

CommanderRoot:
I have rebased the PR on the current master branch.

tacruc:
Hey, thanks would you mind running the replacement over the new files, too?

CommanderRoot:
I did and I also added a new ESLint rule to prevent new usages of `substr`. If you open the PR again you should see the new commit

tacruc:
Actually, I can not. Maybe just create a new PR, sorry.
image

CommanderRoot:
Not a problem. I opened #716

Issue #706, opened 2022-03-24
Removing empty lines from address lookup leads to better results


SeanDS:
The OpenStreetMap Nominatim address lookup seems to function better if empty address lines are removed by Nextcloud Maps prior to building the query string.

Assume an address field for a contact is populated with:

Address: `Copper Beech House Heathside Crescent`
Postcode: `GU22 7BB`
City: `Woking`
Country: `United Kingdom`

The query becomes [`, Copper Beech House Heathside Crescent, Woking, , GU22 7BB, United Kingdom`](https://nominatim.openstreetmap.org/ui/search.html?q=%2C+Copper+Beech+House+Heathside+Crescent%2C+Woking%2C+%2C+GU22+7BB%2C+United+Kingdom) which has no result. If I edit `AddressService.php` around line 171 to add `$splitted_adr = array_filter(array_map('trim', $splitted_adr));`, the query becomes [`Copper Beech House Heathside Crescent, Woking, GU22 7BB, United Kingdom`](https://nominatim.openstreetmap.org/ui/search.html?q=Copper+Beech+House+Heathside+Crescent%2C+Woking%2C+GU22+7BB%2C+United+Kingdom) which returns the correct lookup.

I am happy to make a PR for this if desired.

SeanDS:
Also, somewhat related to this, Nextcloud Maps seems to escape commas inside addresses: address `Copper Beech House, Heathside Crescent` gets queried as `Copper Beech House\, Heathside Crescent, Woking, , GU22 7BB, United Kingdom`. Not sure if this is a bug or whether commas just aren't allowed inside vCard addresses.

SeanDS:
I think the `Extended Address` field should also be moved to the start of the address if it's nonzero, since the spec says this should represent the apartment or suite number (but it's now not recommended to use this field at all).

tacruc:
Hey @SeanDS 
I would be happy to review your PR. If you don't feel like providing a PR do to high, I could apply your suggested code changes.
Best would be if you would provide the code for the function.

SeanDS:
> Hey @SeanDS I would be happy to review your PR. If you don't feel like providing a PR do to high, I could apply your suggested code changes. Best would be if you would provide the code for the function.

Here you go: https://github.com/nextcloud/maps/pull/708

tacruc:
#

PR #708, opened 2022-03-24
Remove blank lines in address before Nominatim query to attempt to improve chance of a match

SeanDS:
See #706.


tacruc (approved):
Cool :+1:

PR #721, opened 2022-04-20
make the photos nav item collapsible

eastyu:


tacruc:
Hi @eastyu,
thanks for your PR. But I think it needs some explanation.
The following questions should be addressed.

- Why?
Photos have no sub item so what should be shown or hidden with the collapsable function.
- Did you test it?
I don't expect this PR to have the desired consequences as the new frontend in the master branch uses the vue navigation component in src/components/.
Therefore I'm not sure whether the navigation template is used at all.

eastyu:
Hi @tacruc ,

please see the screenshot, the text comes over the icon. after I add the class for this item. it shows normal.

image

tacruc:
Hm interesting and this is also happening on the master branch?

eastyu:
I just upgrade to v0.1.10 and the issue is gone. it's fixed somehow. close this PR now.

Issue #723, opened 2022-04-21
Use vue frontend on OpenGeoLink 


tacruc:
Currently the page#opengeolink function still uses the old front end.

tacruc:
#943

PR #724, opened 2022-04-21
Removed old JS frontend

tacruc:
Blocked by #723

@eneiluj  did I miss anything?


tacruc:
@eneiluj When you have time could you check if I missed something?

PR #731, opened 2022-05-01
[Draft] Collaborative Maps Nextgov Hackathon

tacruc:
ToDo:

- [x] apps/maps/maps takes 30 seconds on bigger instances
image
- [ ] This maps branch is incompatible with talk. L.easybutton is not defined if talk app is enabled.
- [ ] Further testing


lib/Controller/FavoritesController.php
```diff
@@ -85,63 +87,162 @@ public function __construct($AppName,
...
+        ],JSON_PRETTY_PRINT);
+    }
+
+    private function getJSONFavoritesFile($folder) {
+        try {
+            $file = $folder->get('.favorites.json');
+        } catch (NotFoundException $e) {
+            $file = $folder->newFile('.favorites.json', $content = $this->defaultFavoritsJSON);
+        }
+        return $file;
```
CarlSchwan, line 102:
```suggestion
    private function getJSONFavoritesFile(OCP\FIles\Folder $folder): OCP\FIles\File {
        try {
            return $folder->get('.favorites.json');
        } catch (NotFoundException $e) {
            return $folder->newFile('.favorites.json', $content = $this->defaultFavoritsJSON);
        }
```

lib/Controller/FavoritesController.php
```diff
@@ -85,63 +87,162 @@ public function __construct($AppName,
...
     }
 
     /**
      * @NoAdminRequired
      */
-    public function getFavorites() {
-        $favorites = $this->favoritesService->getFavoritesFromDB($this->userId);
+    public function getFavorites($myMapId=null) {
```
CarlSchwan, line 108:
```suggestion
    public function getFavorites(?string $myMapId = null): DataResponse {
```

lib/Controller/FavoritesController.php
```diff
@@ -85,63 +87,162 @@ public function __construct($AppName,
...
+			$file = $this->getJSONFavoritesFile($folder);
+			$favoriteIds = $this->favoritesService->addFavoritesToJSON($file, $favorites);
+			$favoritesAfter = [];
+			forEach ($this->favoritesService->getFavoritesFromJSON($file) as $favorite) {
+				if (in_array($favorite['id'],$favoriteIds)) {
+					$favoritesAfter[] = $favorite;
+				}
+			};
+			return new DataResponse($favoritesAfter);
+		}
```
CarlSchwan, line 182:
```suggestion
		}
		$folders = $this->userfolder->getById($myMapId);
		if(!empty($folders) && $this->userfolder->getId() === $myMapId) {
			$folders[] = $this->userfolder;
		}
		$folder = array_shift($folders);
		if(is_null($folder)) {
			return new DataResponse('Map not found', 404);
		}
		$file = $this->getJSONFavoritesFile($folder);
		$favoriteIds = $this->favoritesService->addFavoritesToJSON($file, $favorites);
		$favoritesAfter = [];
		forEach ($this->favoritesService->getFavoritesFromJSON($file) as $favorite) {
			if (in_array($favorite['id'],$favoriteIds)) {
				$favoritesAfter[] = $favorite;
			}
		};
		return new DataResponse($favoritesAfter);
```

lib/Controller/FavoritesController.php
```diff
@@ -85,63 +87,162 @@ public function __construct($AppName,
...
+				) {
+					$this->favoritesService->editFavoriteInJSON($file, $id, $name, $lat, $lng, $category, $comment, $extensions);
+					$editedFavorite = $this->favoritesService->getFavoriteFromJSON($file, $id);
+					return new DataResponse($editedFavorite);
+				} else {
+					return new DataResponse('invalid values', 400);
+				}
+			} else {
+				return new DataResponse('no such favorite', 400);
+			}
```
CarlSchwan, line 221:
```suggestion
				}
				return new DataResponse('invalid values', 400);
			}
			return new DataResponse('no such favorite', 400);
```

lib/Controller/FavoritesController.php
```diff
@@ -151,29 +252,51 @@ public function renameCategories($categories, $newName) {
...
-            return new DataResponse('DELETED');
+    public function deleteFavorite($id, $myMapId=null) {
+        if (is_null($myMapId) || $myMapId === "") {
+            $favorite = $this->favoritesService->getFavoriteFromDB($id, $this->userId);
+            if ($favorite !== null) {
+                $this->favoritesService->deleteFavoriteFromDB($id);
+                return new DataResponse('DELETED');
+            } else {
+                return new DataResponse('no such favorite', 400);
+            }
```
CarlSchwan, line 263:
```suggestion
            }
            return new DataResponse('no such favorite', 400);
```

Also this looks like it's using space instead of tabs here :) Might be a good idea to add php-cs-fix to this repo

lib/Controller/FavoritesController.php
```diff
@@ -151,29 +252,51 @@ public function renameCategories($categories, $newName) {
...
         } else {
-            return new DataResponse('no such favorite', 400);
+            $folders = $this->userfolder->getById($myMapId);
+            $folder = array_shift($folders);
+            $file = $this->getJSONFavoritesFile($folder);
+			if ($this->favoritesService->deleteFavoriteFromJSON($file, $id) > 0) {
+				return new DataResponse('DELETED');
+			} else {
+				return new DataResponse('no such favorite', 400);
+			}
```
CarlSchwan, line 272:
```suggestion
			}
			return new DataResponse('no such favorite', 400);
```

lib/Controller/MyMapsController.php
```diff
@@ -0,0 +1,74 @@
...
+use OCP\IRequest;
+
+class MyMapsController extends Controller {
+
+
+    /* @var MyMapsService */
+    private $myMapsService;
+	private $userId;
+
+    public function __construct($AppName, IRequest $request, MyMapsService $myMapsService, $userId) {
```
CarlSchwan, line 33:
```suggestion
    public function __construct(string $AppName, IRequest $request, MyMapsService $myMapsService, $userId) {
```

lib/Controller/MyMapsController.php
```diff
@@ -0,0 +1,74 @@
...
+        parent::__construct($AppName, $request);
+        $this->myMapsService = $myMapsService;
+		$this->userId = $userId;
+    }
+
+    /**
+     * @NoAdminRequired
+     */
+    public function addMyMap($values) {
```
CarlSchwan, line 42:
```suggestion
    public function addMyMap(array $values): DataResponse {
```

lib/Controller/MyMapsController.php
```diff
@@ -0,0 +1,74 @@
...
+        if (is_string($myMap)) {
+            new DataResponse($myMap, 400);
+        }
+        return new DataResponse($myMap);
+    }
+
+    /**
+     * @NoAdminRequired
+     */
+    public function updateMyMap($id, $values) {
```
CarlSchwan, line 54:
```suggestion
    public function updateMyMap(string $id, array $values): DataResponse {
```

lib/Controller/MyMapsController.php
```diff
@@ -0,0 +1,74 @@
...
+        $myMap = $this->myMapsService->updateMyMap($id, $values, $this->userId);
+        return new DataResponse($myMap);
+    }
+
+    /**
+     * @NoAdminRequired
+     */
+    public function deleteMyMap($id) {
```
CarlSchwan, line 62:
```suggestion
    public function deleteMyMap(string $id): DataResponse {
```

lib/Controller/MyMapsController.php
```diff
@@ -0,0 +1,74 @@
...
+        $result = $this->myMapsService->deleteMyMap($id, $this->userId);
+        return new DataResponse($result);
+    }
+
+    /**
+     * @NoAdminRequired
+     */
+    public function getMyMaps() {
```
CarlSchwan, line 70:
```suggestion
    public function getMyMaps(): DataResponse {
```

appinfo/info.xml
```diff
@@ -33,7 +33,7 @@
     <screenshot>https://raw.githubusercontent.com/nextcloud/maps/master/screenshots/screenshot3.png</screenshot>
     <dependencies>
         <lib>exif</lib>
-        <nextcloud min-version="20" max-version="24"/>
+        <nextcloud min-version="20" max-version="25"/>
```
CarlSchwan, line 36: I would maybe drop 20 and 21 support since it is EOL

@eneiluj what do you think?

CarlSchwan (commented):
Not finished review, it's big!

In general for new code, I would recommend adding type hinting for methods.

tacruc:
@CarlSchwan thanks for the partial review.

For review it might be helpfull to know, that based the branch so that it includes some important bug fixes.

Im on my phone so I cannot give a complete list right now but most are listed in https://github.com/nextcloud/maps/issues/732

lib/Controller/FavoritesController.php
tacruc: For this big code block I would keep the else block

ainola:
It seems that I can't rename a map from the interface (I tried with just hitting [Enter] and also clicking the arrow button). In order to rename "New map" I had to alter the directory name under Maps/

ainola:
I'm also unable to delete a favorite. I can do so on the default map but none of the ones I create.

ainola:
A design critique: Clicking on any of the side items in the map toggles the item on and off. This new feature switches between maps rather than toggles, breaking the consistency of the sidebar. Perhaps this would be better as a drop-down menu or something.

Really nice work!

tacruc:
> It seems that I can't rename a map from the interface (I tried with just hitting [Enter] and also clicking the arrow button). In order to rename "New map" I had to alter the directory name under Maps/

Should be fixed now

tacruc:
> I'm also unable to delete a favorite. I can do so on the default map but none of the ones I create.

Should be fixed too

ainola:
Thanks for all your great work, @tacruc!

tacruc:
> Thanks for all your great work, @tacruc!

Thanks for testing this is really helpful.

JMLatGitHub:
Hi @tacruc, I don't usually ask this sort of question, it makes me really uncomfortable, but my interest in this functionality, especially the trip planning part, is so big that I'm bringing it up nevertheless, always with all the due respect for you and your work: how far are you with the implementation of the trip planning functionality?

Background: We made last year a long trip to the north of Europe, and now we are creating a travel photo book about it. In such cases, usually we copy a map with the itinerary colored by hand into it. We also have another trip coming up in March, this time to the south, for which we start planning in February. I have my own Nextcloud server, and of course it would be very nice if we could use your app for these tasks. The Maps application is installed in the latest version –I mean, the nightly unstable– and at current state I'm able to define new maps, but then can do absolutely nothing with. What I will do later on this weekend is creating a tracking file with the itinerary and trying to read it in. That would be a perfect fallback solution for the photo book task.

It worked without until now, so, absolutely no stress, but if this functionality will be available soon, then we would be enthusiastic and real use case testers... Regardless of that, I take the opportunity here to thank you and others very much for the time and effort that you all invest in the conception and further development of all the fantastic solutions.

Kind regards.

tacruc:
@JMLatGitHub,

> always with all the due respect for you and your work:
Thanks, 

> how far are you with the implementation of the trip planning functionality?

In currently working in public pages, as they are in focus of my personal use case. Having those I hope that I will find some more bugs in order to fix them before the first stable release.

'm a little surprised by one sentence of ours.
>  I'm able to define new maps, but then can do absolutely nothing with. 
You should be able to do everything you could do previously with the maps app. Just now not on one but on multiple maps, which you can share. See screenshots in the readme. What did you expect to do?

JMLatGitHub:
Well, maybe I just don't understand how it works: I have a default map and a Scandinavia map, but since the update (or since the creation of the Scandinavia map, I'm not sure) absolutely no objects are displayed on the maps, no photos, no contacts, no tracks, no devices, which were all visible before and were automatically updated when changes were made, e.g. new photos or contacts... I did already a re-scan but luckless.

By the way, what is the meaning/functionality of public pages?

tacruc:
> since the update
The update to v0.2.5-1-nightly or the update to v0.2.5-2-933-nightly?
I'm just asking as before v0.2.5-1-nightly there shouldn't be any custom map. And if the change to v0.2.5-2-933 breaks something it would be important, as it was rather a small fix, which shouldn't break stuff.

So as I see it you have now two maps, default and Scandinavia.
So it looks like this?
image


Which of the two maps doesn't show content? Both? Or only Scandinavia?

How it should be:
The Default Map should be everything as it used to be.
The Scandinavia map should only show content which you added. (Right click Copy to map) Or by uploading e.g. photos into the related folder. (Probably /Maps/Scandinavia)

tacruc:
> By the way, what is the meaning/functionality of public pages?

Sofar you can share a map like a folder (a map is technically shared by sharing the folder), but only registered and logged in users will be able to view the map. As there is a public page missing which is viewable for not logged in visitors (E.g. people just have a public link to the share).

JMLatGitHub:
Well, I'm pretty sure I had all objects after updating the nightly just before creating the additional Scandinavia map. Now, my setup is as described by you. Additionally, I have My contacts (13), My photos (99+), My tracks (6) and My devices (1) listed, but nothing is shown on none of the two maps.

Thanks for the explanation about the public pages. Nice functionality indeed.

tacruc:
Ok can you check the browser console?
Additionally can you check if the time slider is enabled? Can you disable it?

JMLatGitHub:
Here we go. Slider now disabled. Attached the browser console log accessing Maps: [3rdlvld.2ndlvld.tld-1674484798913.log](https://github.com/nextcloud/maps/files/10480994/3rdlvld.2ndlvld.tld-1674484798913.log).

As I see it, there are errors that are directly related to the Maps application. Beside these, there are also error messages referencing the push service that I have to verify/fix, but these should not affect the functionality of the Maps application, right?

Kind regards

tacruc:
Is it possible for you to temporally disable the talk app and see if that helps?

JMLatGitHub:
Yep, then it's working.

JMLatGitHub:
Another question: For trip planning, especially for longer trips, I think it would be desirable if it would be possible to define individual stages and color them differently or show and hide the individual sections. Also be able to assign different line types or even the possibility to determine the line thickness would also be nice features. Is something like this planned or envisioned?

JMLatGitHub:
Hello again. With the latest nightly map app (V0.2.5-4), I can finally create trips, but only under my admin account. With the normal user accounts, only "Add a favorite", "Place photos", "Place contact" and "Share this location" are available. Is this a known issue/limitation, @tacruc?

JMLatGitHub:
Another issue is that if you export a route, the corresponding GPX route or track is saved with a filename containing the current time in the format of hh:mm:ss. Colon is an unsupported character under Windows and leads to error messages when trying to synchronize.

Best regards

PR #739, opened 2022-06-02
Allways show tracks without timestamps

### LINKED ISSUE #738: Some GPX Tracks not visible
Some GPX tracks dont get rendered on the map even when enabled.

Example GPX File that doesnt work: https://github.com/gps-touring/sample-gpx/blob/master/RoscoffCoastal/1_Roscoff_Morlaix_A_parcours.gpx

Example GPX File that works:
[working.zip](https://github.com/nextcloud/maps/files/8818358/working.zip)

like in my last issue, I‘m pretty sure this worked on an older version.

tacruc:
fixes #738

### DIFF
--- a/src/components/map/TrackLayer.vue
+++ b/src/components/map/TrackLayer.vue
@@ -79,7 +79,7 @@ export default {
 		end: {
 			type: Number,
 			required: false,
-			default: moment.unix(),
+			default: moment().unix(),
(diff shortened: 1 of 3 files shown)


tacruc:
@nougatbyte could you test this?

tacruc:
See #738

PR #771, opened 2022-06-21
Correct text strings

Valdnet:
Signed-off-by: Valdnet <47037905+Valdnet@users.noreply.github.com>

### DIFF
--- a/lib/Controller/DevicesController.php
+++ b/lib/Controller/DevicesController.php
@@ -215,11 +215,11 @@ class DevicesController extends Controller {
                 return new DataResponse($this->l->t('/Maps is not a directory'), 400);
             }
             else if (!$mapsFolder->isCreatable()) {
-                return new DataResponse($this->l->t('/Maps is not writeable'), 400);
+                return new DataResponse($this->l->t('/Maps directory is not writeable'), 400);
             }
         }
         else {
-            return new DataResponse($this->l->t('Impossible to create /Maps'), 400);
+            return new DataResponse($this->l->t('Impossible to create /Maps directory'), 400);
         }
 
         $nbDevices = $this->devicesService->countPoints($this->userId, $deviceIdList, $begin, $end);
--- a/src/components/map/ClickPopup.vue
+++ b/src/components/map/ClickPopup.vue
@@ -59,7 +59,7 @@
 
 			<div v-if="allowEdits" class="buttons">
 				<button class="primary" @click="handleAddToFavorites">
-					{{ t("maps", "Add to Favorites") }}
+					{{ t("maps", "Add to favorites") }}
 				</button>
 			</div>
 		</template>
--- a/lib/Controller/FavoritesController.php
+++ b/lib/Controller/FavoritesController.php
@@ -269,12 +269,12 @@ class FavoritesController extends Controller {
                 return $response;
             }
             else if (!$mapsFolder->isCreatable()) {
-                $response = new DataResponse($this->l->t('/Maps is not writeable'), 400);
+                $response = new DataResponse($this->l->t('/Maps directory is not writeable'), 400);
                 return $response;
             }
         }
         else {
-            $response = new DataResponse($this->l->t('Impossible to create /Maps'), 400);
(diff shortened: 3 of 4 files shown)


src/components/map/ClickPopup.vue
```diff
@@ -59,7 +59,7 @@
 
 			<div v-if="allowEdits" class="buttons">
 				<button class="primary" @click="handleAddToFavorites">
-					{{ t("maps", "Add to Favorites") }}
+					{{ t("maps", "Add to favorites") }}
```
rakekniven, line 62: "Favorites" same as "Files"

lib/Controller/DevicesController.php
```diff
@@ -215,11 +215,11 @@ public function exportDevices($deviceIdList, $begin, $end, bool $all=false): Dat
                 return new DataResponse($this->l->t('/Maps is not a directory'), 400);
             }
             else if (!$mapsFolder->isCreatable()) {
-                return new DataResponse($this->l->t('/Maps is not writeable'), 400);
+                return new DataResponse($this->l->t('/Maps directory is not writeable'), 400);
```
rakekniven, line 218: Why not using "Folder"?

rakekniven (commented):
See my comments

lib/Controller/DevicesController.php
tacruc: "Folder" or "folder"?

rakekniven: folder

src/components/map/ClickPopup.vue
Valdnet: "Files" is the name of the application. This string appears in two other buttons and is described differently.

https://github.com/nextcloud/maps/blob/9e04e711620ca42672029db88116b078feac1310/src/components/map/ClickSearchPopup.vue#L17
https://github.com/nextcloud/maps/blob/9e04e711620ca42672029db88116b078feac1310/src/components/map/PoiMarker.vue#L17

lib/Controller/DevicesController.php
Valdnet: @tacruc I would also point out this issue #772.

lib/Controller/DevicesController.php
```diff
@@ -215,11 +215,11 @@ public function exportDevices($deviceIdList, $begin, $end, bool $all=false): Dat
...
             }
         }
         else {
-            return new DataResponse($this->l->t('Impossible to create /Maps'), 400);
+            return new DataResponse($this->l->t('Impossible to create /Maps directory'), 400);
```
tacruc, line 222: What do you think about quoting the path?
```suggestion
            return new DataResponse($this->l->t('Impossible to create "/Maps" directory'), 400);
```

tacruc (commented):
Thanks, looks like an improvement already

lib/Controller/DevicesController.php
Valdnet: I thought about that as well, and it should be correct, but the code needs to be corrected because after quoting the path, a backslashw is displayed in the message.

Example: `"Impossible to create \"/Maps\" directory"`

In general, you should remove the leading and trailing quotes in your code. In my opinion, they are unnecessary. All you need is a colon.

Example: `Failed to import devices: Impossible to create "/Maps" directory`

tacruc: > In general, you should remove the leading and trailing quotes in your code. In my opinion, they are unnecessary. All you need is a colon.

I don't understand this. Isn't your example exactly what I wrote. 

What I get is that we have a rendering problem.

Valdnet: > I don't understand this. Isn't your example exactly what I wrote.

You do not need to put quotation marks after the colon, as this only adds to the less legibility of the message. See an example screenshot:

obraz

I suggest removing the quotation marks in all messages. This will improve readability, especially when you want to put the path in quotation marks.

tacruc: Now, I see what your talking about.

rakekniven (approved):

PR #781, opened 2022-07-11
[WIP] php8.1 and min version nc22

tacruc:
PHP 8.1 support


tacruc:
@eneiluj do you think we should merge this before the release v0.2.0 on Monday, or after?

PR #798, opened 2022-07-25
fixes mapbox no panning, by downgrading mapbox-gl to 2.4.1

### LINKED ISSUE #786: Mapbox tiles are not moved
Hello all.

I just updated the map to version 0.2.0 on my Nextcloud 24.0.2 (php 8.0.21).

Now when I move the map (street map) with the mouse, the map stops and I only move the pictures, contacts or routes.
It looks like I move all layers above the map.

Can anyone confirm this?

Greeting Helix

tacruc:
There is an Issue with mapbox-gl-leaflet in combination with mapbox-gl v2.5.1 https://github.com/mapbox/mapbox-gl-leaflet/issues/145 
Fixes #786.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
 				"leaflet.featuregroup.subgroup": "^1.0.2",
 				"leaflet.locatecontrol": "^0.76.1",
 				"lrm-graphhopper": "^1.3.0",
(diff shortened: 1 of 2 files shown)


eneiluj (approved):
Nice, thankfully it works with previous versions of mapbox-gl

Issue #674, opened 2021-12-13
show only recent history for my devices


mortee:
The "My Devices" feature is awesome, I'm very happy I can use it instead of e.g. G location history. That's why I have PhoneTrack constantly sending my position to my server.

However, the usability of the track quickly becomes quite limited, as it starts to show several days, weeks, months' worth of movements. Would it be possible to set a customizable time cutoff for the track display? Of course, that shouldn't mean the data points themselves should be deleted, just optionally only the recent ones shown.

stefanoginella:
+1 that would be immensely useful!

tacruc:
#970 is tested on tracks with 1M points I think this is the bedder fix.

mortee:
The automatic switch to heatmap beyond a certain number of points is definitely a cool feature.

However, I fail to see how it replaces (with a "bedder fix") the ability to view only my recent track, which would obviously provide a different kind of information from my all-time heatmap, and also be actually useful as opposed to the track of e.g. several years' worth of movements within a city.

So, while the heatmap function is definitely welcome, I still wish I could see my recent actual track.

mortee:
I wasn't even aware of the time filter feature. If that indeed does what I would assume, that would be a good solution for what I would like to achieve.

However, now that I activated it, apparently I can go back with it only a couple of minutes into the past, I can't drag the left slider any further.

![kép](https://user-images.githubusercontent.com/338540/220419440-8a05df10-da4a-4ce4-87c3-6b2283fc7e03.png)

tacruc:
Ok some Tips:
- It should get orange at some points (at the edge, or when the sliders are close together) when you let go there the scale should adapt
- The range is by default the min and the max timpstamp of elements shown, If nothing is activated it might by from now to now
- Double clicking should reset to the default timespan
- There might be a bug (but I think I fixed it, that the range doesn't change, if you set the slider on the edge)

mortee:
Well, for me, the bar turns yellow once or twice, and it adapts itself to let more room at the end, but this only lets me go back a few minutes, and then the slider stays at the bar's end.

My track goes back years, so it takes a lot of time to load - and when it finally appears, the bar doesn't update to accomodate it all.

NC 25.0.3, Firefox 110 on Windows.

PR #970, opened 2023-02-12
Devices heatmap

tacruc:
Creates Nice Heat-maps for devices histories. They are a way faster and nice then rendering the Track. They are shown for histories with more then 2500 points.

image

Enabling and disabling devices is reasonably fast. There is still a noticeable freezing on the app, within the getDevice function in App.vue.

### DIFF
--- a/src/views/App.vue
+++ b/src/views/App.vue
@@ -225,7 +225,7 @@ import AppNavigationDevicesItem from '../components/AppNavigationDevicesItem'
 import AppNavigationMyMapsItem from '../components/AppNavigationMyMapsItem'
 import optionsController from '../optionsController'
 import { getLetterColor, hslToRgb, Timer, getDeviceInfoFromUserAgent2, isComputer, isPhone } from '../utils'
-import {binSearch, getToken, isPublic} from '../utils/common'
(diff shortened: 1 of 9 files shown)


src/views/App.vue
```diff
@@ -2002,12 +2012,13 @@ export default {
...
 		disableDevice(device) {
 			device.enabled = false
+			device.historyEnabled = false
 			this.saveEnabledDevices()
 		},
 		getDevice(device, enable = false, save = true, zoom = false) {
 			device.loading = true
 			network.getDevice(device.id, this.myMapId).then((response) => {
-				this.$set(device, 'points', response.data.sort((p1, p2) => (p1.timestamp || 0) - (p2.timestamp || 0)))
+				this.$set(device, 'points', response.data /*.sort((p1, p2) => (p1.timestamp || 0) - (p2.timestamp || 0))*/)
```
tacruc, line 2021: @julien-nc  Do you know, if this line can freeze the browser for a noticeable time, if points is a big array? Notice that the sorting is disabled?

tacruc:
Fixes  #674, as long histories won't break the browser

PR #984, opened 2023-03-01
[Blocked] php8.2 wait for dropping php 8.0 support

tacruc:


tacruc:
We have to support php 8.0 for a while longer unfortunately

PR #989, opened 2023-03-03
Add action to report errors

### LINKED ISSUE #884: Submitting OSM map (error) notes
There exist similar third party mapping projects that also support [adding POI](https://wiki.openstreetmap.org/wiki/Onosm.org) or reporting errors via forms over map notes. Consider whether such a simple workflow of contribution could be integrated to the map widget.

https://wiki.openstreetmap.org/wiki/Notes

See also #883

tacruc:
Fixes  #884

image

### DIFF
--- a/src/report-error-map-action.js
+++ b/src/report-error-map-action.js
@@ -1,0 +1,9 @@
+
+OCA.Maps.registerMapsAction({
+	label: t('maps', 'Report Error'),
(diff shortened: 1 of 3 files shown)


tacruc:
@bkil what do you think?

bkil:
Splendid. I would probably phrase it as `Report map error`, though.

Issue #1013, opened 2023-03-22
Add an option to deduplicate same address for a contact


vincowl:
I have some contacts who have the same professional address as their personal one. It would be great to add an option to deduplicate these adresses, thus it would display only one marker instead of multiple ones for a same contact. If it sounds interesting I can try to submit a PR

vincowl:
Fine ! I'll donthat and submit a PR as soon as I'll have Time for this. I'll juste deduplicate if addresses are exactly the same string.

tacruc:
Fix thanks to @vincowl

PR #1024, opened 2023-03-27
Deduplicate contact's adresses

vincowl:
Aims to solve #1013


tacruc:
Hi @vincowl,
thanks for your PR and your contribution. I really appreciate it.  If I understand this correctly you always compare it with the previous address.

I see two cases where this procedure might fail:
- White spaces or different order of parts of the address, which still map to the same coordinates
- If a contact has 3 addresses and addresses 1 and 3, I think your deduplication won't detect it.

Therefore I would suggest:
- doing the deduplication on the level of $geo attributes. E.g deduplicate if the distance is approximately smaller than 5m. 
- keeping an array of previous $geo to be able to duplicate with respect to all previous addresses of the contact.

- Merging the address type of all deduplicated addresses


What do you think? If you need help feel free to ask.

vincowl:
True. I just had the same thought once I submitted this but decided to wait for your opinion. Seems perfect to me. In my case, there was no chance that address strings could be different, that's why I didn't use $geo.

I will create a class attribute holding the precision used to deduplicate.

What do you mean by "Merging the address type" ? Keeping the first one, concatening the strings or something else ?

tacruc:
I thought It would be cool to somehow see all address types in the contact popup shown at the map.

The easiest way would be to just concatenate the strings. 
A nicer way would be to provide a array and then think about how to visualize the array the nicest way in the frontend code.

Sorry I'm short on time right now to think about the details. You can implement it if you have a good idea.

vincowl:
@tacruc I think this should be something close to your thoughts. Let me know how to make it better.

tacruc:
Thanks for improving, I see you went all the way including frontend changes🥳.
I'm on vacation and only have a phone with me. Will do the full review when I have my laptop back next week.

vincowl:
No problem ! There's no rush : enjoy your vacation ! Nothing's above having a good reste 😉

vincowl:
@tacruc Hope you had a great vacation. Any news on this PR ?

tacruc:
@vincowl thanks, vacations was great. Sorry, the PR slipped. I will have a look this evening.

tacruc:
Hi @vincowl 
thanks, seams to work nicely. I like how multiple address type tags are shown.
Is there a particular reason, why you committed the changed composer.lock file? I think this is making Issues with the tests.

tacruc:
Actually, what happens now on the 'delete' this address button? Probably we should hide it on markers with multiple addresses right?

vincowl:
> Hi @vincowl 
> thanks, seams to work nicely. I like how multiple address type tags are shown.
> Is there a particular reason, why you committed the changed composer.lock file? I think this is making Issues with the tests.

Oops.... I'll remove it !

vincowl:
> Actually, what happens now on the 'delete' this address button? Probably we should hide it on markers with multiple addresses right?

I have not the slightest idea how this behaves... I'll better have a look at this and test before deciding what to implement or hide or upgrade

vincowl:
> Actually, what happens now on the 'delete' this address button? Probably we should hide it on markers with multiple addresses right?

I have a trouble here... I cannot delete any address wheter it is duplicated or not and whatever the addressbook it belongs to. I've got an error 400 saying READONLY. Any idea what may cause this ?

tacruc:
```php
@@ -239,6 +239,9 @@ class ContactsController extends Controller {
                                     } else {
                                         // Concatenate AddressType to the corresponding record
                                         array_push($result[$prevRes[$duplicatedIndex]]['ADRTYPE'], $adrtype);
+                                                                               $result[$prevRes[$duplicatedIndex]]['isUpdateable'] = false;
+                                                                               $result[$prevRes[$duplicatedIndex]]['isDeletable'] = false;
+                                                                               $result[$prevRes[$duplicatedIndex]]['isShareable'] = false;
                                     }
                                 }
```
You could add the following lines.

tacruc:
The issue for the readonly is that we compare an int to str in line 647
`			if ($book['id'] === $bookid) {`

tacruc:
Additionally a test in
https://github.com/nextcloud/maps/blob/7e87ae5a69cfb95804f7de72f8317e846e64f18f/tests/Unit/Controller/ContactsControllerTest.php
https://github.com/vincowl/nextcloud-maps/blob/7e87ae5a69cfb95804f7de72f8317e846e64f18f/tests/Unit/Controller/ContactsControllerTest.php
would be extremely nice.

vincowl:
> if ($book['id'] === $bookid) {

OK. This is solved now. I just modified it to 
`if ($book['id'] === (int) $bookid) {` since the CardDav backend ensures that `$book['id']` is always an int and your functions forces its input to be a string. I think its better than just using a `==`

tacruc:
Maybe forcing the input to be a string is not the best idea then, or?

vincowl:
> ```
> @@ -239,6 +239,9 @@ class ContactsController extends Controller {
>                                      } else {
>                                          // Concatenate AddressType to the corresponding record
>                                          array_push($result[$prevRes[$duplicatedIndex]]['ADRTYPE'], $adrtype);
> +                                                                               $result[$prevRes[$duplicatedIndex]]['isUpdateable'] = false;
> +                                                                               $result[$prevRes[$duplicatedIndex]]['isDeletable'] = false;
> +                                                                               $result[$prevRes[$duplicatedIndex]]['isShareable'] = false;
>                                      }
>                                  }
> ```
> 
> You could add the following lines.
> 
> Is the easy option. But a delete button next to the address type is definitely cool too.

Now that I can test the deletion, it appears that the actual code deletes the first address in the "duplicates" list. This may not be what we want since, one could be willing to delete the n-th one instead of the first. 

For instance, let's imagine I have a contact named Bob who is a freelance developer and also an art seller, who works remotely and owns 2 companies (one for the dev and one for the art part). So he will have 3 addresses that are exactly the same. Bob shares its contact card with me but I don't care about it's dev company. So I will remove this address which would be linked with the corresponding label. Removing the first address forces me to edit its contact and modify every label attached to its addresses.

I definitely think we need to add the trash button next to the labels.... Hope it's not too difficult !

tacruc:
Next to the labels would be Ideal. But just making the marker readonly, e.g. by setting isUpdateable and isDeletable to false hides the button, which would be fine for me a first implementation.

vincowl:
> Next to the labels would be Ideal. But just making the marker readonly, e.g. by setting isUpdateable and isDeletable to false hides the button, which would be fine for me a first implementation.

Fine ! The trash thing could be done in a second step which would allow to end this PR and open a new one

vincowl:
> tests/Unit/Controller/ContactsControllerTest.php

No problem. I'll just check how to do this and run the tests. Should not be too tricky !

vincowl:
@tacruc can't run the unit tests through composer. My nextcloud instance is deployed via a docker container and maps app is in a volume linked to my file system where I can code. But running `composer run test:unit` sends an error 
```
Error in bootstrap script: Error:
Failed opening required '/srv/docker/prod/containers/nextcloud/custom_apps/maps/tests/../../../tests/bootstrap.php' (include_path='.:/usr/share/php')
```
Which `bootstrap.php` is required on line 13 of maps/tests/bootstrap.php ?

vincowl:
@tacruc Could you please tell me how to run the unit tests please ? I have no `tests` folder at the root of my nextcloud instance. Is there a dedicated instance of nextcloud to deploy or a command to run to get this folder and its `bootstrap.php` file ?

composer.lock
tacruc, line 1: Why is removed if not tracked?

vincowl: I added it to git by mistake and pushed it. So I had to run a `git rm` and all tests ran perfectly then. That's why it is shown as deleted. 
Now I added it locally to my own list of files to ignore to ensure that it would not be pushed anymore. That's why I wrote that it is not tracked. I should have added "not tracked anymore"

tacruc:
I usually use this
https://github.com/juliushaertl/nextcloud-docker-dev
for my development setup. 
The test folder are in the nextcloud/server git repository. It might be that this directory is not included in the release bundles.  

Does this help or is related to you problem: https://help.nextcloud.com/t/bootstrapping-nextcloud-for-phpunit/135165

vincowl:
> I usually use this https://github.com/juliushaertl/nextcloud-docker-dev for my development setup. The test folder are in the nextcloud/server git repository. It might be that this directory is not included in the release bundles.
> 
> Does this help or is related to you problem: https://help.nextcloud.com/t/bootstrapping-nextcloud-for-phpunit/135165

OK. Now I can run tests and add my own ones. But I have strictly no idea how to add a contact to the database, set its addresses and then retrieves the infos. Unit tests regarding contacts only test the value of the request's status code. No contacts are created and there are only TODOS not implemented ... Do you have any infos about this or where to find the related doc ?

tacruc:
@vincowl I gave not forgotten you. When I find a minute I will write a test. And then we can merge. Sorry for the delay.

tacruc:
@vincowl I gave not forgotten you. When I find a minute I will write a test. And then we can merge. Sorry for the delay.

tacruc:
@vincowl I have not forgotten you. When I find a minute I will write a test. And then we can merge. Sorry for the delay.

vincowl:
No problem at all ! I also have a very few time left on personal projects and this is not a blocking thing tonmy usage. So no need to hurry. Thanks for this kind message 😉

vincowl:
No problem at all ! I also have a very few time left on personal projects and this is not a blocking thing tonmy usage. So no need to hurry. Thanks for this kind message 😉

vincowl:
@tacruc If it's easier for you, just give me a link to a test in an other project you would have heard about or docs about contacts testing.

vincowl:
I think everything should be here to add/check the contacts, right ? [https://github.com/nextcloud/server/blob/master/lib/public/Contacts/IManager.php](https://github.com/nextcloud/server/blob/master/lib/public/Contacts/IManager.php)

tacruc:
Hi @vincowl I just updated your PR, but I coundn't push to your branch as I didn't have the permissions, so I made a new one #1072.

vincowl:
Hi @tacruc 
You should have received an invitation to my repository so you can work directly in this branch. Maybe it would be easier to keep this feed. I merged the current diffs between this branch and the master one as you did in #1072

2 remarks :
- Maybe I am wrong but there is no new test implemented
- Running unit tests leads to errors like `" General error: 1 no such table: oc_maps_tracks" `or `" General error: 1 no such table: oc_maps_favorites"`

tacruc:
Merged with #1072, Thanks @vincowl again, for your contribution.

vincowl:
That's fine !!! Thanks @tacruc !!

Issue #923, opened 2023-01-08
[Bug]: Track/Route filename contains a colon, which causes problems on Android


zukaitis:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Maps app **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

When creating a track or route file, default filename contains at least one colon, which makes the file not downloadable/unopenable on Android (and possibly other OSes)

### Steps to reproduce

1. Create a track
2. Try to open it on Android device

### Expected behavior

Unclear errors, renaming the file in Nextcloud fixes this

### Installation method

Manual installation

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install of the server?

Fresh Nextcloud Server install

### Is this bug present after an update or on a fresh install of the app?

Fresh Nextcloud Maps install (never installed before)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### List of activated Apps

```shell
Enabled:
  - accessibility: 1.10.0
  - activity: 2.16.0
  - bookmarks: 11.0.4
  - bruteforcesettings: 2.4.0
  - calendar: 3.5.4
  - circles: 24.0.1
  - cloud_federation_api: 1.7.0
  - comments: 1.14.0
  - contacts: 4.2.3
  - contactsinteraction: 1.5.0
  - cookbook: 0.9.18
  - dashboard: 7.4.0
  - dav: 1.22.0
  - federatedfilesharing: 1.14.0
  - federation: 1.14.0
  - files: 1.19.0
  - files_pdfviewer: 2.5.0
  - files_rightclick: 1.3.0
[...]

PR #1034, opened 2023-04-21
Removes all colons from generated track/route filename

zukaitis:
Apparently, #923 wasn't fully fixed before, because 'replace()` only changes single occurrence in the string. Therefore, violating colon currently still appears in the filename. This PR should hopefully fix it for good


tacruc:
Thanks

PR #1039, opened 2023-04-25
Support HEIC format for command scan-photos

benstonezhang:


tacruc:
Did this work for you?

benstonezhang:
Seem the result not store in db, there must be more logic to handle this. :-(

benstonezhang:
I do can print the lat and lon in log for heic files, but not sure why they not stored in db

tacruc:
This is interesting. 
Can you send me the heic photo which is working for you?

benstonezhang:
the `file` command output show this heic file use JPEG instead of HEVC:
`
JPEG image data, Exif standard: [TIFF image data, big-endian, direntries=14, manufacturer=Apple, model=xxxxxxxxx, orientation=upper-left, xresolution=216, yresolution=224, resolutionunit=2, software=15.4.1, datetime=xxxxxxxxx, hostcomputer=xxxxxxxxx, baseline, precision 8, 3854x2449, components 3
`
So the bug [77757](https://bugs.php.net/bug.php?id=77757) is really still not fixed.

PR #1083, opened 2023-06-30
Add location field to favorites form

alaskanpuffin:
Adds a location field to the favorites form containing coordinates of the point. The position can also be edited by manually changing the coordinates.

image


tacruc (approved):
Nice 👍

PR #1110, opened 2023-08-06
Fix nominatim search address query

Simounet:
Hi there,
Searching an address with a `/` is [not allowed anymore](https://github.com/osm-search/Nominatim/issues/3134). Here is a fix with the new format.
Have a nice day.


tacruc (approved):
👍 and thabks

PR #1295, opened 2024-07-13
Update OSM tile service URL

### LINKED ISSUE #127: Optimize loading of tiles from OSM?
Hello,

sometimes the map tiles are loaded really slow which results in white spots or blurred squares on the map. I don't have this problem with GpxPod/Motion/Phonetrack also build by @eneiluj on the same server. I can see several GET requests to different tile servers in the browser console and it takes a long time (more than a minute at times) to get them from the server.

Is there anything that can be done to speed up the loading time of the tiles? I don't think it is a problem with the connection speed because I tried it on an internet connection with up to a gigabit.

StyXman:
Using URL from https://operations.osmfoundation.org/policies/tiles/

Closes #127.


jancborchardt (approved):
Seems good to me as per the docs, but not tested. @julien-nc @tacruc :)

julien-nc (approved):

PR #1336, opened 2024-10-06
Support Nextcloud 30, update CSP header

### LINKED ISSUE #1333: Please make the app compatible with NC30
It would be great to see it working on the newest NC version!

Ma27:
Closes #1333 
It probably makes most sense to review this commit-by-commit.

cc @come-nc @susnux @tacruc


css/style.scss
```diff
@@ -575,7 +575,7 @@ tr.selected td {
     text-align: center;
 }
 .tooltip-photo-name {
-    color: var(--color-text-light);
+    color: var( --color-main-text);
```
come-nc, line 578:
```suggestion
    color: var(--color-main-text);
```

lib/Controller/ContactsController.php
```diff
@@ -679,7 +679,7 @@ private function setAddressCoordinates(float $lat, float $lng, string $adr, stri
 		$req = $qb->execute();
 		$result = $req->fetchAll();
 		$req->closeCursor();
-		$qb = $qb->resetQueryParts();
+		$qb = $this->dbconnection->getQueryBuilder();
```
come-nc, line 682: The `$qb` property should be removed from all classes, they should instead only get a query builder from the connection when they need it.

lib/Controller/ContactsController.php
```diff
@@ -702,7 +702,8 @@ private function setAddressCoordinates(float $lat, float $lng, string $adr, stri
 				]);
 			$req = $qb->execute();
 			$id = $qb->getLastInsertId();
-		}$qb = $qb->resetQueryParts();
+		}
+		$qb = $this->dbconnection->getQueryBuilder();
```
come-nc, line 706:
```suggestion
```
There is no point of doing that without using the object after?

tests/bootstrap.php
```diff
@@ -8,7 +8,6 @@
  * @author Julien Veyssier <eneiluj@posteo.net>
  * @copyright Julien Veyssier 2019
  */
-define('PHPUNIT_RUN', 1);
```
come-nc, line 11: Why?

lib/Controller/ContactsController.php
Ma27: Let me re-check later, but IIRC this was necessary because `$qb` is a reference to `$this->qb` (not sure if it applies to this class, but I had a few other classes where this was the case) and not resetting this caused some internal errors in the query builder's code down the line.

But you have a point that `$qb` should be local to a method and not a property of a class, so with that we could definitely get rid of this.

Ma27:
Btw I've seen that the CI failed because of a missing Signed-off-by, will fix it up on the next iteration.

AndyScherzinger:
cc @julien-nc I guess

lib/Service/AddressService.php
```diff
@@ -149,18 +147,18 @@ private function lookupAddressInternal($adr): array {
 
 		$adr_norm = strtolower(preg_replace('/\s+/', '', $adr));
 
-		$this->qb->select('lat', 'lng')
+		$qb->select('lat', 'lng')
```
come-nc, line 150: You forgot to assign $qb in this method

come-nc (commented):
All good apart from this one forgotten getQueryBuilder call.

Ideally calls to execute should be replaced with executeQuery/executeStatement but that’s not removed in 30 yet so it’s unrelated to this PR.

come-nc (approved):

tacruc:
Hi @Ma27,
Thanks a lot for your contribution I appreciate it a lot. 
Unfortunately I'm currently quite busy. Right now I'm sitting in a train and wanted to use the time to update and merge this. Unfortunately my laptop run out of battery and the plugs are not working.
So I'm stuck with my phone.

Could you update the Test workflow such that the tests are run on NC30?
https://github.com/nextcloud/maps/blob/master/.github/workflows/test.yml

Ma27:
Hi @tacruc!
No worries at all!

Pushed a commit that does that. Not sure if it's completely correct though, I usually work with other CI systems :upside_down_face: 
However, I think you should be able to push to my fork's branch as well, right?

tacruc:
Thanks a lot. In theory I should be able to push to your branch. In practice I didn't find a good git client for Android. Do you know one?

Can you remove v28 and v29 from testing and make the DCO, happy? I think then we can go for a merge.

Ma27:
Done.

tacruc (approved):

tacruc:
@Ma27, there is a nightly build with your PR https://apps.nextcloud.com/apps/maps

PR #1363, opened 2025-01-01
Fix tiles not working on public pages because of missing CSP entries

chefe:
A CSP entry for https://tile.openstreetmap.org was introduced in 8e9e960ac7bf8c9e7ac638775504b007242bb3e9, but it was only added to the PageController but it should be added to all the places where the CSP headers are generated.

cc @tacruc, @Ma27


tacruc:
@chefe good catch thanks.

tacruc (approved):

PR #1375, opened 2025-01-16
Fix wrong failure-msg when clearing photo-cache.

umgfoin:
`bool GeophotoService::clearCache` returns `false`, if any of the cleared caches was already empty. 
Accordingly, a failure-msg is shown in the maps-GUI. 
This change makes the function always return `true `unless an exception occurred - then `false `is returned.


tacruc:
@come-nc What do you think? I would prefere having the Icache->clear() funktion fixed. But this is far out of our code space. So I'm fine.

Issue #1392, opened 2025-02-26
Nextcloud 31


RainerEmrich:
Please add compatibility for Nextcloud 31.

alexanderhuzar:
Bump.

Ma27:
With all due respect, [bumping issues is considered harmful](https://sindresorhus.com/blog/issue-bumping).

This is only causing notification spam for people like me who subscribed the issue to get updates on that matter.

rcfa:
Is it known to be broken with 31, or is it just not tested?

RainerEmrich:
> Is it known to be broken with 31, or is it just not tested?

I just did a short test which looked good so far. This i with nc 31.0.1.

xcellentavi:
> > Is it known to be broken with 31, or is it just not tested?
> 
> I just did a short test which looked good so far. This is with nc 31.0.1. So, it should be easy to add compatibility with nc 31.

That's at least good to know. I'm going to stick with NC 30 until this is updated, but hopefully @tacruc or someone else can let us know if they have an update soon or recommend upgrading even though it is untested.

johansmitsnl:
Is there an update on the support for NC31 or what needs to be done?

RainerEmrich:
> Is there an update on the support for NC31 or what needs to be done?

I think a new version is on its way, see the latest commits.

tacruc:
There is a nightly release, haven't had time to do much testing. Feel free to provide feedback. If it is similarly stable as the other versions we can publish it.

duck-rh:
Since there is still `min-version="30"` declared I wanted to upgrade the app to 1.6.0-3-nightly first to avoid disabling the app before upgrading to NC31; unfortunately that did not go well and the whole site is broken:

```
[...]

Issue #1384, opened 2025-02-09
PHP 8.4


fr3ddy-sudo:
Please add Support for the PHP 8.4 Version

rcfa:
Is it known to break with 8.4, or is it simply not tested and/or throws a few warnings?

fr3ddy-sudo:
Es ist ein Fehler bei der Anfrage aufgetreten. Es kann nicht fortgefahren werden.
App "Maps" cannot be installed because the following dependencies are not fulfilled: PHP with a version lower than 8.3 is required

umgfoin:
> Is it known to break with 8.4, or is it simply not tested and/or throws a few warnings?

Generally, I see neither critical nor deprecation issues with PHP8.4 (I'm currently running a minimally patched rel. under 8.4).
However PHP8.4 is supported no earlier as of NC31, thus I'd suggest to include these changes with a NC31 ready release as requested in #1392 .

jaro3:
https://dev.to/gromnan/fix-php-84-deprecation-implicitly-marking-parameter-as-nullable-is-deprecated-the-explicit-nullable-type-must-be-used-instead-5gp3

utility fixes the code....

umgfoin:
Not exactly:
Your first contribution will fix the code.

umgfoin:
PHP 8.4 support implemented with rel 1.6.0 (#1431).

PR #1402, opened 2025-04-12
Update Dependencies and Nextcloud Requirement

a4blue:
This should make the Maps App Compatible with Nextcloud 31, it worked locally on a minimal Nextcloud Installation.
In order to achieve this i updated some dependencies (especially the Tools in vendor-bin) and re-ran cs:fix and also psalm.


a4blue:
#1392 #1384 Should be fixed by this

PR #1418, opened 2025-05-20
chore(tests): Cleanup bootstrap.php to be forward-compatible

come-nc:
This uses the new tests/autoload.php from nextcloud/server instead of messing directly with OC private autoloader.

See https://github.com/nextcloud/server/pull/52951 and https://github.com/nextcloud/server/pull/52945 for context.

I did not add a direct reference to `tests/autoload.php` for this app because `tests/bootstrap.php` is required which requires it indirectly.


provokateurin:
CI seems to be stuck

susnux:
> CI seems to be stuck

I think there is no such workflow anymore but still setup as required

tacruc (approved):

PR #1464, opened 2025-09-23
fix: use filepicker API for placing photos on the map

susnux:
- [x] :warning:  based on #1462 

With Nextcloud 28 the `OC.dialogs` API was deprecated, instead we need to use `@nextcloud/dialogs` for the filepicker. So this fixes:
1. The deprecated `OC.dialogs` usage
2. Bug where both dialogs where opened at the same time (file and folder picker).
3. Properly place photos on other than the default map


tacruc (approved):
Hi Sunex thanks for the PR.

PR #1465, opened 2025-09-25
AddressService: Use the structured query when possible

### LINKED ISSUE #712: Extended addresses prevent matches for contact map location
Thanks for maintaining this!

Many of my contacts have apartment numbers or some other reason to use the "extended address" field. Once these contacts have this field set, however, their icon does not show up on the map. Removing the extended address allows them to show up.

For example:

150 West 95th Street / Apt 1 / New York / NY / 10025
^This does not show up

150 West 95th Street / New York / NY / 10025
^This does show up

Would it make sense to just not pass in the extended address for showing contact locations, or would that break something else?

tintou:
This allows nominatim to make sure that we are requesting the right place and enables more heuristics on their side.
Fixes #712

### DIFF
--- a/lib/Service/AddressService.php
+++ b/lib/Service/AddressService.php
@@ -181,16 +181,28 @@ class AddressService {
 
 			// we get rid of "post office box" field
 			$splitted_adr = explode(';', $adr);
-			if (count($splitted_adr) > 2) {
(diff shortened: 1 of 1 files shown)


tintou:
Without this, none f my addresses were showing in Maps, with this I can see all of them

PR #1497, opened 2026-02-05
refactor(Controller): Modernize code

CarlSchwan:
- Use getFirstNodeById for performance reason
- Add typing
- Use property contrustor promotion
- Use attributes instead of annotations


susnux (approved):
Nice!
Tests sre related though. Tests need to be adjusted

joshtrichards:
If `FilesHooks` is still around when you wrap this up, I'll take care of replacing the legacy hooks once it's merged (similar to how I just did nextcloud/notes#1836). ;-)

PR #1632, opened 2026-07-23
Fix search box

sim6:
- Remove text "label" on top of search box.
- Pan map when the search dropdown item is selected.
- Keep search dropdown open to show search results.


sim6:

CarlSchwan (approved):

sim6:

PR #1633, opened 2026-07-23
fix: eliminate 404 error when clicking on the map

sim6:
The ClickSearchPopup marker icon was created with L.icon({ iconUrl: 'noIcon' }), a relative URL that resolved to /apps/maps/noIcon which does not exist on the server.

Replace with L.divIcon producing a zero-size invisible element. The marker only serves as an anchor for the popup and does not need a visual icon, so a div element with no dimensions is sufficient.


sim6:

CarlSchwan (approved):

sim6:
:)

PR #1636, opened 2026-07-23
fix: remove favorite marker from map immediately after deletion

sim6:
Replace `this.$delete()` with native `delete` in all favorite deletion paths (single delete, bulk delete, category unlink).

`this.$delete()` is a Vue 2 compatibility shim that may not reliably trigger Vue 3's Proxy-based reactivity system when removing keys from a reactive object. This caused the FavoritesLayer deep watcher on displayedFavorites to not fire, leaving stale markers on the map until a full reload.

In Vue 3, the native `delete` operator is intercepted by the Proxy's deleteProperty trap, which properly triggers reactivity.

Affected methods:
- onFavoriteDelete (single favorite via sidebar/popup)
- onFavoritesDelete (bulk delete via cluster right-click)
- onDeleteFavoriteCategoryFromMap (unlink shared category)

### DIFF
--- a/src/views/App.vue
+++ b/src/views/App.vue
@@ -1615,7 +1615,7 @@ export default {
 				this.selectedFavorite = null
 				this.closeSidebar()
 				showError(t('maps', 'Favorite was deleted'))
-				this.$delete(this.favorites, favid)
+				delete this.favorites[favid]
 			}).catch((error) => {
 				console.error(error)
 			})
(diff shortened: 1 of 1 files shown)


sim6:

CarlSchwan (approved):

CarlSchwan:
Thanks for the fix and sorry for the delay, so much stuff to do lately on my $work time :(

sim6:

sim6:
@CarlSchwan No rush at all, and thanks for your work on this.

PR #1641, opened 2026-07-24
fix: switch to correct map when clicking 'My maps' sidebar entry

sim6:
Without an emits declaration, Vue 3 treats `@click` on AppNavigationMyMapItem as a native DOM event listener, causing a raw MouseEvent to reach loadMap instead of the myMap object.


sim6:

CarlSchwan (approved):

sim6:

CarlSchwan:
Not needed anymore

PR #1645, opened 2026-07-26
fix: prevent popup from reopening after adding favorite from map click

sim6:
Add @click.stop modifier to the 'Add to favorites' button click handler in ClickSearchPopup to call event.stopPropagation(). Without it, the click event bubbles up through the Leaflet popup overlay to the map container, triggering onMapClick and opening a duplicate popup.


sim6:

CarlSchwan (approved):

sim6:
:)

